PR #19, opened 2016-07-21
Save image link

### LINKED ISSUE #2: Save Image Link
#### As a user I want to save a link to an image.

Acceptance criteria:
- [ ] The link is entered through a form.
- [ ] After the form is saved, the link is persisted in the database.
- [ ] After the form is submitted, the user is redirected to a page displaying the image.

Dependencies:
- Deployment Environment

__Note__: Many image sharing sites will permit one to upload images, or
download images when provided a link. For this project we will only work with
the link, i.e., URL, to various images.

lsq921224:
Closes #2


app/controllers/images_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def show
+    @image = Image.find(params[:id])
+  end
+
+  private
+
+  def image_params
+    params.require(:image).permit(:link)
```
leoi11: good use of strong parameters. It helps eliminates security issues!

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  def index
+    @images = Image.all
+  end
+
+  def new
+    @image = Image.new
+  end
+
+  def create
+    @image = Image.new(params[:image].permit(:link))
```
leoi11: it looks like you have created a `image_params` below, and you might want to use that because `params[:image]` doesn't behave the same as `params.require(:image)`

app/views/images/index.html.erb
```diff
@@ -0,0 +1,15 @@
+<form action="/images/new">
```
leoi11: Consider using path helpers in rails. It's a good practice to avoid hardcoding paths anywhere as it will involve a lot of effort to change a path.

leoi11: I notice that this form lives in index.html.erb.
This should not be the case. In rails, this kind of forms to create a new resource should live in new.html.erb.

lsq921224: This is only meant to be a link to /images/new.. should I change it to <a>?

app/views/images/index.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <input type="submit" value="new image">
+</form>
+<table>
```
leoi11: What you have from this line to the end of the file is the correct use of the index page. (Display all resources in one page.) However, I don't see it as a acceptance criteria for issue #2. Is there any reasons you made this change in this PR?

app/views/images/index.html.erb
leoi11: yes. Please use the rails `link_to` helper to generate links :)

leoi11: And for some reason I thought it's a GET request, which is incorrect to say.
The default method is POST for form submit.
So we shouldn't be submitting a form to `images/new` with a POST method anyways.

app/views/images/index.html.erb
lsq921224: I did it because I think it's a good way to verify whether all the image links are added correctly, but it's not in this acceptance criteria. I will remove them from this PR.

app/views/images/new.html.erb
```diff
@@ -0,0 +1,10 @@
+<p>Add New Image</p>
+<%= form_for @image do |f| %>
+    <p>
```
leoi11: Is there any reasons to have a <p> tag here? What are you trying to address here?

lsq921224: There is no particular reason. It's just my habit to wrap text in html with some sort of tags, so later if I want to change the style of the text, it would be easier. Do you mean that I should use some other tag like h1,h2 or I should not use any tag here?

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
-  root 'welcome#index'
+  resources :images
```
leoi11: This rails keyword `resources` generates 7 actions, and some of them we don't really have.
It's a good practice to use `:only` in this case to limit it to only the actions the controller have.

app/views/images/new.html.erb
leoi11: What happens when there are no tags at all? If we need something in the future, we should add those in the future.
In general, it's good to avoid adding things ahead of time unless we know it's going to happen in very near future.

leoi11:
I haven't finished reviewing yet. Please feel free to make the changes first, and I'll come back as soon as I have time.

lsq921224:
Thank you; I have made corresponding changes according to your comments.

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    @image = Image.new(image_params)
+    if @image.save
+      redirect_to @image, alert: 'Image link created successfully.'
+    else
+      render 'new'
+    end
+  end
+
+  def show
+    @image = Image.find(params[:id])
```
leoi11: Please consider adding some error handling when image id cannot be found.

app/views/images/index.html.erb
```diff
@@ -0,0 +1 @@
+<%= link_to "New Image", new_image_path %>
```
leoi11: thanks for using the path helpers, looks perfect!

app/controllers/images_controller.rb
lsq921224: For now if the image id cannot be found, I will render the index page, which only contains a link to :new

app/models/image.rb
```diff
@@ -0,0 +1,3 @@
+class Image < ApplicationRecord
+  validates :link, presence: true
```
HaominZhang: You can also use `validates_presence_of :link` here, which is more common in our APM code base

HaominZhang:
I think we can add a flow test case to test the saving image link flow. Have you attended any session about how to write flow tests?

test/models/image_test.rb
```diff
@@ -0,0 +1,8 @@
+require 'test_helper'
+
+class ImageTest < ActiveSupport::TestCase
+  def test_link_not_nil
+    image = Image.new
+    assert_not image.save, 'Saved the image without a link'
+  end
+end
```
HaominZhang: How about adding another test for save image link success?

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    follow_redirect!
+    assert_response :success
+    assert_select 'img[src=?]', @image.link
+  end
+
+  test 'should get show' do
+    get image_path(@image)
+    assert_response :success
+  end
+end
```
HaominZhang: Can we add another test case where the link cannot be saved?

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    assert_difference 'Image.count', 1 do
+      post images_path, image: { link: @image.link }
+    end
+
+    assert_response :redirect
+    follow_redirect!
+    assert_response :success
+    assert_select 'img[src=?]', @image.link
+  end
+
```
HaominZhang: Can we add another test case where the link cannot be saved?

lsq921224:
is flow test integration test? I attended 'intro to testing' which briefly introduced integration test, but we did not have the chance to actually write one. I can try to write one by copying and pasting http://guides.rubyonrails.org/testing.html#integration-testing

test/controllers/images_controller_test.rb
lsq921224: sure, will do!

HaominZhang:
In our flow test, we will use a gem called `Capybara`. I believe you will have a separate session learning how to write flow test, so don't worry now we can come back and work on it after you learn it.

app/models/image.rb
rmacklin: @HaominZhang It's more common in our codebase because we have a lot of legacy code and this used to be the only way to do validations. Using regular `validates` is fine, and encouraged.

rmacklin:
There is no session for plain flow tests - there's only a session for refactoring flow tests to use page objects. To create a flow test, add an `images_crud_test.rb` file in `tests/flows/` which defines an `ImagesCrudTest` class that inherits from [`FlowTestCase`](https://github.com/AppFolioOnboarding/image-sharer-lsq921224/blob/a64c940afb846d6e57c322177af03e130b79bed8/test/flow_test_helper.rb#L7). Then use the [capybara](https://github.com/jnicklas/capybara) documentation to write a test that goes through the flow of adding a new image.

app/models/image.rb
lsq921224: @HaominZhang @rmacklin when writing tests, i notice there are two ways of declaring them: one is 
```
def test_link_cannot_be_nil
    image = Image.new
    assert_not image.save, 'Saved the image without a link'
end
```
the other is
```
test 'should get index' do
    get images_path
    assert_response :success
end
```
Is there any differences between these two? which one should I pick?

rmacklin: I prefer the latter because it allows you do be more expressive - the test case can be more "human readable" (you can use spaces, apostrophes, etc.). In APM, we mostly use the former because originally that was the only way to declare tests.

Gemfile.lock
```diff
@@ -114,7 +114,7 @@ GEM
     pronto-rubocop (0.6.2)
       pronto (~> 0.6.0)
       rubocop (~> 0.38, >= 0.35.0)
-    puma (3.4.0)
+    puma (3.5.2)
```
rmacklin: Why were all of these gems upgraded?

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+      redirect_to @image, alert: 'Image link created successfully.'
+    else
+      render 'new', status: :unprocessable_entity
+    end
+  end
+
+  def show
+    @image = Image.find(params[:id])
+  rescue ActiveRecord::RecordNotFound
+    render 'index', status: :not_found
```
rmacklin: `rescue` should be used to handle _exceptional_ behavior. In this case, we don't need to use `rescue` if we just use `Image.find_by(id: params[:id])` instead. That will return either the image or `nil`, so you can just write an `if` `else` statement to handle the cases of whether the image exists or not.

app/views/images/new.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= form_for @image do |f| %>
+    <%= f.label 'new image link' %><br />
+    <%= f.text_field :link %>
+    <%= f.submit 'Add' %>
```
rmacklin: Minor, but you should configure your editor to use 2 spaces (rather than 4) for indentation

Gemfile.lock
lsq921224: Because I did a 'bundle update', is that something that shouldn't be done/pushed?

circle.yml
```diff
@@ -0,0 +1,3 @@
+test:
+  post:
+    - bundle exec rubocop
```
rmacklin: This change is from https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/20. It shouldn't be part of this PR. You should rebase your branch off the latest master to fix that.

test/fixtures/images.yml
```diff
@@ -0,0 +1,2 @@
+sample_image:
+  link: 'http://www.w3schools.com/css/trolltunga.jpg'
```
rmacklin: We actually avoid using fixtures for almost everything, because they end up doing more harm than good. For instance, if you have this fixture file, you cannot write a test where there are no `Image`s without first calling `Image.destroy_all`. Additionally, setting up test data as part of your individual test cases improves their readability - you don't have to refer to some other file to see what'll be in the database for a given test.

rmacklin: You'll still want to commit a fixture file though - just leave it empty which ensures your images table will be cleared out between tests (which is important for test isolation)

test/models/image_test.rb
```diff
@@ -0,0 +1,8 @@
+require 'test_helper'
+
+class ImageTest < ActiveSupport::TestCase
+  def test_link_not_nil
+    image = Image.new
+    assert_not image.save, 'Saved the image without a link'
```
rmacklin: It'd be better to make assertions about the validity directly, rather than asserting on the return value of `.save`. You can use:
```ruby
assert_predicate image, :invalid?
```
to assert this more directly. Additionally, you should test that the error message for the invalid image is what you expect it to be.

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,33 @@
...
+  test 'should get show' do
+    get image_path(@image)
+    assert_response :success
```
rmacklin: In general, controller tests should verify something about the content of the rendered page using `assert_select`. In this case, we want to verify that the show page displays the correct image.

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,33 @@
+require 'test_helper'
+
+class ImagesControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    @image = images(:sample_image)
+  end
```
rmacklin: Do we need this setup?

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,33 @@
...
+    assert_response :success
+  end
+
+  test 'should be able to create' do
+    assert_difference 'Image.count', 1 do
+      post images_path, image: { link: @image.link }
+    end
+
+    assert_response :redirect
+    follow_redirect!
```
rmacklin: Even though this works in Rails 5, we prefer to test only a single action per controller test case. You can change this to
```ruby
assert_redirected_to image_path(Image.last)
```

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  test 'should get index' do
+    get images_path
+    assert_response :success
+  end
+
+  test 'should get new' do
+    get new_image_path
+    assert_response :success
```
rmacklin: As in https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/19#discussion_r72190518, you should add a basic assertion about the content of the page.

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,33 @@
+require 'test_helper'
+
+class ImagesControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    @image = images(:sample_image)
+  end
+
+  test 'should get index' do
+    get images_path
+    assert_response :success
```
rmacklin: As in https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/19#discussion_r72190518, you should add a basic assertion about the content of the page.

db/schema.rb
```diff
@@ -0,0 +1,21 @@
...
+# system, you should be using db:schema:load, not running all the migrations
+# from scratch. The latter is a flawed and unsustainable approach (the more migrations
+# you'll amass, the slower it'll run and the greater likelihood for issues).
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 20160721165007) do
+
+  create_table "images", force: :cascade do |t|
+    t.string   "link", null: false
```
rmacklin: Did you manually modify this to have `null: false`? It's not specified in the [migration](https://github.com/AppFolioOnboarding/image-sharer-lsq921224/blob/760b429aac58666c17375712779d0dcb41fd6305/db/migrate/20160721165007_create_images.rb#L4)

app/controllers/images_controller.rb
```diff
@@ -0,0 +1,30 @@
+class ImagesController < ApplicationController
+  def index
+    @images = Image.all
```
rmacklin: You aren't using this, so it shouldn't be part of this PR

Gemfile.lock
rmacklin: Well, it's okay to do that but that change is not necessary to complete the "save image link" story. So it should be done on its own branch, in order to keep the scope of the changes in each branch as small as possible. In a real world application, each gem upgrade require regression testing to make sure no bugs were introduced, so that's why it's especially important to limit the scope of the changes of each branch.

db/schema.rb
lsq921224: Yes. I did not read the comments in this file and thought I did the right thing.

rmacklin: No worries! It's a learning experience :)

So you'll want to run `rails db:rollback` to roll back the migration, then edit the migration to include `null: false` for the `link` attribute, and then re-run `rails db:migrate` which should regenerate the schema with `null: false` set automatically. Once you've verified that, you'll want to squash that change into the commit that originally created the migration so that the schema and migrations are never out of sync.

app/controllers/images_controller.rb
lsq921224: well, but I will be using this soon enough. Is it that important that we only make changes on what the issue requires?

rmacklin: This is a bit of an extreme case since it's only one line, but generally yes you want to implement only what's necessary for the story you're working on. One issue with doing it like this is that when you make a PR for the images index story this won't appear in the changeset for that story which makes it easier to overlook (again this is an extreme case since its just one line, so there aren't edge cases to be aware of, but you get the idea)

test/flows/images_crud_test.rb
```diff
@@ -0,0 +1,12 @@
...
+class ImagesCrudTest < FlowTestCase
+  test 'add new image' do
+    visit(new_image_path)
+    fill_in('image[link]', with: 'http://www.w3schools.com/css/trolltunga.jpg')
+    find('input[type="submit"]').click
+    assert has_current_path?(image_path(Image.last))
+    assert_equal 'Image created successfully.', find('.notice').text
+    assert_equal Image.last.link, find('img')[:src]
+  end
+end
```
HaominZhang: How do you think if we add a flow before what you've done, in which we directly save an empty link, and we can assert if we get to the right page, and then come back again to save a valid link?

HaominZhang:
It is really great that you refactored the code to make the test and implementation go in the same commit!

test/flows/images_crud_test.rb
lsq921224: That sounds like a plan.

HaominZhang:
Looks good to me!

app/views/images/show.html.erb
```diff
@@ -0,0 +1,4 @@
+<% flash.each do |name, msg| -%>
+    <%= content_tag :div, msg, class: name %>
+<% end -%>
```
bboe: This is a bit redundant to the flash on the new page. Can you DRY this up? Is there a shared _layout_ that this can be added to?

bboe:
@lsq921224 I made just one comment that I'd like you to address prior to merging. Separately I would like to say really great work putting the individual changes into atomic commits.

test/flows/images_crud_test.rb
```diff
@@ -0,0 +1,16 @@
...
+    visit(new_image_path)
+    fill_in('image[link]', with: '')
+    find('input[type="submit"]').click
+    assert has_current_path?(images_path), "Does not have #{images_path}"
+    assert_equal 'Failed to create image.', find('.error').text
+    fill_in('image[link]', with: 'http://www.w3schools.com/css/trolltunga.jpg')
+    find('input[type="submit"]').click
+    assert has_current_path?(image_path(Image.last)), "Does not have #{image_path(Image.last)}"
+    assert_equal 'Image created successfully.', find('.notice').text
+    assert_equal Image.last.link, find('img')[:src]
```
rmacklin: It would be better to use `'http://www.w3schools.com/css/trolltunga.jpg'` for the expected value here. The reason is that this assertion could pass if both `Image.last.link` and `find('img')[:src]`, but that would mean your application had a bug where it wasn't saving the links properly (so we'd want this assertion to fail if such a a bug existed).

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,39 @@
+require 'test_helper'
+
+class ImagesControllerTest < ActionDispatch::IntegrationTest
+  test 'should get new' do
+    get new_image_path
+    assert_response :success
+    assert_select 'form[action=?]', images_path
+  end
+
+  test 'should be able to create' do
```
rmacklin: How about `'should be able to create a valid image'`?

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,39 @@
...
+    assert_difference 'Image.count' do
+      post images_path, params: { image: { link: 'http://www.w3schools.com/css/trolltunga.jpg' } }
+    end
+    assert_equal 'Image created successfully.', flash[:notice]
+    assert_redirected_to Image.last
+  end
+
+  test 'should not be able to create' do
```
rmacklin: And `'should not be able to create an invalid image'`?

test/controllers/images_controller_test.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  test 'should get show' do
+    image = Image.create!(link: 'http://www.w3schools.com/css/trolltunga.jpg')
+    get image_path(image)
+    assert_response :success
+    assert_select 'img[src=?]', image.link
+  end
+
+  test 'should fail to show invalid image' do
```
rmacklin: Can we call this `'should fail to show a nonexistent image'` (that is, replace _invalid_ with _nonexistent_ to be more accurate?

PR #21, opened 2016-07-28
Image index

### LINKED ISSUE #3: Image Index
#### As a user I want the homepage to display all saved images.

Acceptance criteria:
- [ ] New images that are added show up on the homepage.
- [ ] These images are persisted if the browser is closed or even if the
  server is restarted.
- [ ] Images are not displayed wider than 400px.

Dependencies:
- Save Image Link

lsq921224:
Closes #3 
#23 #24 are based on this branch


test/flows/images_crud_test.rb
```diff
@@ -14,4 +14,27 @@ class ImagesCrudTest < FlowTestCase
...
+    find('input[type="submit"]').click
+    visit(images_path)
+    assert page.has_css?('img', count: 1), 'Home page should have one image'
+    assert_equal link, find('img')[:src]
+  end
+
+  test 'images on homepage should have class of .img_medium_display' do
+    link = 'http://www.w3schools.com/css/trolltunga.jpg'
+    visit(new_image_path)
+    fill_in('image[link]', with: link)
```
aaronpanch: You could seed the database with an image already, instead of "filling" out the form. This is fine too :)

rmacklin: This test can be done with a controller test of the index action by using `assert_select` and it would be much faster than a flow test.

PR #22, opened 2016-07-29
Image url validation

### LINKED ISSUE #4: Image URL Validation
#### As a user I want to be prevented from submitting invalid image URLs.

Acceptance criteria:
- [ ] I cannot successfully save an image with an invalid URL.
- [ ] An error message is associated with the appropriate input field on
  failure.

Discussion Topic:
- [ ] What is a valid image URL?

Dependencies:
- Save Image Link

lsq921224:
Closes #4


app/models/image.rb
```diff
@@ -1,3 +1,31 @@
...
+      uri_is_image_url(uri)
+    end
+  rescue URI::InvalidURIError
+    errors.add(:link, 'is not a valid uri')
+  rescue URI::InvalidComponentError
+    errors.add(:link, 'contains unrecognized uri component')
+  end
+
+  def uri_is_image_url(uri)
+    errors.add(:link, 'must starts with either http or https') unless
```
bboe: This two line _unless_ statement at the end of the command is very hard to read. Can we refactor that?

app/models/image.rb
```diff
@@ -1,3 +1,31 @@
+require 'uri'
+
 class Image < ApplicationRecord
   validates :link, presence: true
+  validate :image_url_is_valid
+
+  private
+
+  IMAGE_EXTENTIONS = %w(.jpg .jpeg .png .gif).freeze
```
bboe: [Pedantic] Can we sort this list? Keeping lists like this sorted means we're less likely to introduce duplicates.

app/models/image.rb
```diff
@@ -1,3 +1,31 @@
...
+  PROTOCOLS = %w(http https).freeze
+
+  def image_url_is_valid
+    uri = URI.parse(link)
+    if uri.host.nil?
+      errors.add(:link, 'is not valid url: no host')
```
bboe: Do your users require this level of precision in the URL validation? If they do great, but I will expect to see a test for each message type.

app/models/image.rb
```diff
@@ -1,3 +1,31 @@
...
+    else
+      uri_is_image_url(uri)
+    end
+  rescue URI::InvalidURIError
```
bboe: If we don't require separate messages can we simplify this code?

app/views/images/new.html.erb
```diff
@@ -1,5 +1,10 @@
 <%= form_for @image do |f| %>
   <%= f.label 'new image link' %><br />
   <%= f.text_field :link %>
+    <% if @image.errors.full_messages.any? %>
```
bboe: Is this conditional necessary?

test/flows/images_crud_test.rb
```diff
@@ -8,6 +8,14 @@ class ImagesCrudTest < FlowTestCase
     find('input[type="submit"]').click
     assert has_current_path?(images_path), "Does not have #{images_path}"
     assert_equal 'Failed to create image.', find('.error').text
+    url_error_msgs = page.all('.url_error_msg')
+    assert_equal "Link can't be blank", url_error_msgs[0].text
```
bboe: We have to be really careful with direct assertions in capybara tests. Such assertions assume the object under test has already loaded on the page. If you reorder some of the assertion statements around, this such tests can become flakey.

Consider using the `has_` set of assertions: http://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Matchers

test/flows/images_crud_test.rb
```diff
@@ -8,6 +8,14 @@ class ImagesCrudTest < FlowTestCase
...
+    assert_equal 'Link is not valid url: no host', url_error_msgs[1].text
+    fill_in('image[link]', with: 'http://google.com/invalid.go')
+    find('input[type="submit"]').click
+    assert has_current_path?(images_path), "Does not have #{images_path}"
```
bboe: I really like that you're adding a message to these plain assertion statements.

test/models/image_test.rb
```diff
@@ -11,4 +11,67 @@ class ImageTest < ActiveSupport::TestCase
     image = Image.new(link: 'http://www.w3schools.com/css/trolltunga.jpg')
     assert_predicate image, :valid?
   end
+
+  test 'invalid urls should all fail' do
+    invalid_urls = Array[]
```
bboe: Is there a simpler way to create this array of urls?

test/models/image_test.rb
```diff
@@ -11,4 +11,67 @@ class ImageTest < ActiveSupport::TestCase
...
+    invalid_urls.push('')
+    invalid_urls.push('invalid')
+    invalid_urls.push('inva lid')
+    invalid_urls.push('in va lid')
```
bboe: What's the difference between this test case and the one above it?

test/models/image_test.rb
```diff
@@ -11,4 +11,67 @@ class ImageTest < ActiveSupport::TestCase
...
+    invalid_urls.push('in va lid.com/a.jpg')
+    invalid_urls.push('invalid.com')
+    invalid_urls.push('invalid.com/not.img')
+    invalid_urls.push('http:invalid')
```
bboe: Is this necessary when you have `http:invalid.jpg`?

test/models/image_test.rb
```diff
@@ -11,4 +11,67 @@ class ImageTest < ActiveSupport::TestCase
...
+    invalid_urls.push('http:invalid.jpg')
+    invalid_urls.push('http:/invalid.jpg')
+    invalid_urls.push('http:/invalid.jpg/in.jpg')
+    invalid_urls.push('http:// invalid.com/in.jpg')
+    invalid_urls.push('http://invalid.com/ in.jpg')
+    invalid_urls.push('http://invalid')
+    invalid_urls.push('https://invalid')
+    invalid_urls.push('https://invalid.com')
+    invalid_urls.push('https://invalid.com/not.img')
+    invalid_urls.push('https://invalid.com/notimg')
```
bboe: Should the suffix of this be `notjpg`? Are you testing to ensure the `dot` is there?

test/models/image_test.rb
```diff
@@ -11,4 +11,67 @@ class ImageTest < ActiveSupport::TestCase
...
+    invalid_urls.push('https://invalid.com/jpg')
+    invalid_urls.push('https://invalid.com/a.jpgpng')
+    invalid_urls.push('https://invalid.jpg')
+    invalid_urls.push('ftp://invalid.com/not.png')
+    invalid_urls.each { |url| assert_predicate Image.new(link: url), :invalid? }
```
bboe: What can you assert about the error message?

test/models/image_test.rb
```diff
@@ -11,4 +11,67 @@ class ImageTest < ActiveSupport::TestCase
...
+  end
+
+  test 'invalid urls should generate corresponding error messages' do
```
bboe: Oh.. I see you have the tests for the specific error messages here. If you want to maintain separate error messages, then each error message should have its own test function rather than combining them all into one test.

test/models/image_test.rb
```diff
@@ -11,4 +11,67 @@ class ImageTest < ActiveSupport::TestCase
...
+    assert_equal 'must starts with either http or https', image.errors.messages[:link].first
+    image = Image.new(link: 'mailto:invalid.com/a.jpp')
+    assert_predicate image, :invalid?
+    assert_equal 'contains unrecognized uri component', image.errors.messages[:link].first
+    image = Image.new(link: '//invalid.org/scheme-relative/URI/with/absolute/path/to/invalid.txt')
+    assert_predicate image, :invalid?
+    assert_equal 'must starts with either http or https', image.errors.messages[:link][0]
+    assert_equal 'is not an image url', image.errors.messages[:link][1]
+  end
 end
```
bboe: Where are all the valid URL tests? We'd like to see those to ensure that what is valid today doesn't become invalid in the future if someone makes a mistake in modifying the code.

bboe:
Added my initial code review. I will like to look at it again once it has been updated. While you're at the initial set of reviews feel free to rebase to get it up-to-date.

test/flows/images_crud_test.rb
```diff
@@ -7,12 +7,19 @@ class ImagesCrudTest < FlowTestCase
     fill_in('image[link]', with: '')
     find('input[type="submit"]').click
     assert has_current_path?(images_path), "Does not have #{images_path}"
-    assert_equal 'Failed to create image.', find('.error').text
+    assert has_css?('.url_error_msg', count: 2), 'There should be two error messages.'
+    assert has_content?("Link can't be blank"), "There should be error message: Link can't be blank"
```
lsq921224: I wanted to do something like`assert has_css?('.url_error_msg:nth-child(1)', text:"Link can't be blank")`, but unfortunately it didn't work.

bboe: I believe you can use `find` to first find the element, and then call `has_content` on that element.

bboe: Try using all('.url_error_message')[1].has_text. Something along those lines should work.

lsq921224: Yes, this works

PR #24, opened 2016-08-01
Delete image

### LINKED ISSUE #8: Delete Image
#### As a user I want to delete an image.

Acceptance criteria:
- [x] The user is prompted for confirmation prior to deletion.
- [x] Deleted images are removed from the homepage.

Dependencies:
- Image Index

lsq921224:
Closes #8 
This branch is based on #21


test/controllers/images_controller_test.rb
```diff
@@ -51,4 +51,25 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    assert_no_difference 'Image.count' do
+      delete image_path(-1)
+    end
+    assert_equal 'Failed to delete image.', flash[:error]
+    assert_redirected_to images_path
+  end
+
+  test 'should be able to delete valid image' do
+    image = nil
+    assert_difference 'Image.count' do
```
D-D-M: This `assert_difference` is basically testing the `create!` method on `Image`. The important assertion here is on [line 68](https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/24/files#diff-8807b315bee3099305dbdee3d4f0a4f7R68), and that will definitely fail if `Image.create!` fails, so this assertion is unnecessary/superfluous.

test/flows/images_crud_test.rb
```diff
@@ -37,4 +37,37 @@ class ImagesCrudTest < FlowTestCase
...
+    link = 'http://www.w3schools.com/css/trolltunga.jpg'
+    visit(new_image_path)
+    fill_in('image[link]', with: link)
+    find('input[type="submit"]').click
+    visit(images_path)
+    assert page.has_css?('a[data-method="delete"]', count: 1), 'Delete link should appear'
+  end
+
+  test 'delete a valid image' do
+    Capybara.current_driver = :selenium
```
D-D-M: If you'd like to have this functionality for all of your flow tests (which would be a good idea), you can add this line to your `flow_test_helper.rb`. An example of that is done [here](https://github.com/AppFolioOnboarding/image-sharer-wpliao1989/pull/25/files#diff-e32affa1c612d33896f619dfcede19cfL30). The default javascript driver is selenium, so
```
Capybara.current_driver = :selenium
```
and
```
Capybara.current_driver = Capybara.javascript_driver
```
basically do the same thing, but the latter is slightly less brittle.

test/controllers/images_controller_test.rb
lsq921224: Yes, that's what I did. But then my assert_difference for deletion will not give me what I want. It will return 0 instead of -1 because I created an image then deleted it (it's comparing 0 with 0). I hope you know what I mean.. I think maybe I should put both the create and delete inside the assert_difference block.

test/flows/images_crud_test.rb
```diff
@@ -37,4 +37,37 @@ class ImagesCrudTest < FlowTestCase
...
+    the_other_link = 'http://www.fakeimage.com/fake.jpg'
+    visit(new_image_path)
+    fill_in('image[link]', with: the_other_link)
+    find('input[type="submit"]').click
+    visit(new_image_path)
+    fill_in('image[link]', with: link)
+    find('input[type="submit"]').click
+    visit(images_path)
+    find("a[href='/images/#{Image.last.id}']").click
+    page.driver.browser.switch_to.alert.dismiss
```
D-D-M: Capybara has some nifty `dismiss_confirm`, `accept_confirm`, and `click_link` methods, their signature looks something like this:
```
dismiss_confirm('Are you sure?') do
  click_link('Delete Image or Whatever Your Delete Button Text Says')
end
```

test/flows/images_crud_test.rb
lsq921224: Oh, I didn't know Capybara.javascript_driver at all. I will try this out, maybe by using this I can avoid updating the selenium webdriver as well.

test/flows/images_crud_test.rb
```diff
@@ -37,4 +37,37 @@ class ImagesCrudTest < FlowTestCase
...
+    assert has_current_path?(images_path), "Does not have #{images_path}"
+    assert page.has_css?("img[src='#{link}']", count: 1), 'Image should still exist'
+    find("a[href='/images/#{Image.last.id}']").click
+    page.driver.browser.switch_to.alert.accept
```
D-D-M: Again, I would use `accept_confirm` here. I think `click_link` also takes a selector as an argument, so you don't need to pass the text of the button here if you're trying to delete one of two images. But that raises a question: Why does this test need two images in the first place? Seems like we can still test the delete functionality with just one image. The benefit of this would be that you don't have to specify which image you're talking about (`Image.last`) when you're trying to pick which one to delete.

D-D-M:
Other than the few comments I had, this looks pretty good!

test/flows/images_crud_test.rb
rmacklin: Actually I think it's a good thing if your flow test verifies that the _correct_ image was deleted when you delete an image.

PR #26, opened 2016-08-03
Image index

lsq921224:
Merge delete_index to master


app/controllers/images_controller.rb
```diff
@@ -21,6 +21,17 @@ def show
...
   end
 
+  def destroy
+    @image = Image.find_by(id: params[:id])
+    if @image.present?
+      @image.destroy!
+      redirect_to images_path, notice: 'Image deleted successfully.'
+    else
+      flash.now[:error] = 'Failed to delete image.'
+      redirect_to images_path
```
bboe: Does this error message convey to the user what it should? It didn't actually fail to delete because such an image did not exist to delete.

In what cases would a user legitimately reach this error message? And what do you think they would want to happen in those cases? It might be of benefit to discuss this offline.

test/flows/images_crud_test.rb
```diff
@@ -37,4 +37,36 @@ class ImagesCrudTest < FlowTestCase
     assert page.has_css?('img', count: 1), 'Home page should have one image'
     assert page.has_css?('.img_medium_display'), 'Images on home page should have .img_medium_display property'
   end
+
+  test 'delete link should only appear when there are images' do
```
bboe: Is this a flow test?

PR #29, opened 2016-08-04
Add 'respond_to' to image controllers

lsq921224:
> By adding `respond_to` the response status is 406 when attempting to request an invalid content type, whereas normally rails returns a 500 error.


bboe:
I'm going to close this, since it isn't actually needed to avoid the 500 exception in rails 5.

PR #30, opened 2016-08-04
Image tags

### LINKED ISSUE #7: Image Tags
#### As a user I want to add tags to images.

Acceptance criteria:
- [ ] When I add a new image, there is a form field to add tags.
- [ ] After I've saved an image with tags, I see those tags displayed with the
  associated image on the index page.
- [ ] My application uses the acts-as-taggable-on gem to provide the tagging
  functionality.

Reference material:
- https://github.com/mbleigh/acts-as-taggable-on

Dependencies:
- Image Index

lsq921224:
Closes #7


db/schema.rb
```diff
@@ -10,12 +10,29 @@
...
+  create_table "taggings", force: :cascade do |t|
+    t.integer  "tag_id",                    null: false
+    t.string   "taggable_type",             null: false
+    t.integer  "taggable_id",               null: false
+    t.string   "tagger_type"
+    t.integer  "tagger_id"
+    t.string   "context",       limit: 128, null: false
+    t.datetime "created_at",                null: false
+    t.index ["tag_id"], name: "index_taggings_on_tag_id", unique: true
+    t.index ["taggable_id", "taggable_type", "context"], name: "index_taggings_on_taggable_id_and_taggable_type_and_context"
```
shfung: Did you use `rake acts_as_taggable_on_engine:install:migrations` to install all the migrations?

test/controllers/images_controller_test.rb
```diff
@@ -5,14 +5,15 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     get new_image_path
     assert_response :success
     assert_select 'form[action=?]', images_path
+    assert_select 'input#image_tag_list'
   end
 
   test 'should be able to create a valid image' do
     assert_difference 'Image.count' do
       post images_path, params: { image: { link: 'http://www.w3schools.com/css/trolltunga.jpg' } }
```
shfung: Can you add `tag_list` as one of the params?

db/schema.rb
lsq921224: Yes, then I removed 5 of the 6 generated migration files because I thought they were not necessary.

app/views/images/new.html.erb
```diff
@@ -1,8 +1,11 @@
 <%= form_for @image do |f| %>
-  <%= f.label 'new image link' %><br />
+  <%= f.label 'new image link' %>
   <%= f.text_field :link %>
     <% @image.errors.full_messages.each do |error_message| %>
       <%= content_tag :div, error_message, class: 'url_error_msg' %>
     <% end %>
+    <br />
+    <%= f.label 'tags' %>
+    <%= f.text_field :tag_list %>
```
shfung: Are the tags still comma separated if you have submitted the form with an invalid URL and got back an error?

shfung:
You should also add controller tests and unit tests for the model.

db/schema.rb
shfung: I'm wondering why do you think they are not necessary?

lsq921224: well, please correct me if my thoughts are wrong. the 5 files are:

1. add_missing_taggable_index.
```
  def self.up
    add_index :taggings, [:taggable_id, :taggable_type, :context]
  end

  def self.down
    remove_index :taggings, [:taggable_id, :taggable_type, :context]
  end
```
This is basically addressed by
```
  def change
    ...
    add_index :taggings, [:taggable_id, :taggable_type, :context]
  end
```
in the file that I have

2.add_taggings_counter_cache_to_tags
This migration adds taggings_count column to tags, I added this column in
```
  def create_tags_table
    create_table :tags do |t|
      t.string :name, null: false
      t.integer :taggings_count, default: 0, null: false
    end
  end
```

3.change_collation_for_tag_name
This executes a SQL if the db is mysql, for which I'm not using

4.add_missing_indexes
```
  def change
    add_index :taggings, :tag_id
    add_index :taggings, :taggable_id
    add_index :taggings, :taggable_type
    add_index :taggings, :tagger_id
    add_index :taggings, :context

    add_index :taggings, [:tagger_id, :tagger_type]
    add_index :taggings, [:taggable_id, :taggable_type, :tagger_id, :context], name: 'taggings_idy'
  end
```
This just adds extra indexes, and I don't think they are very necessary

5. add_missing_unique_indices
```
  def self.up
    add_index :tags, :name, unique: true

    remove_index :taggings, :tag_id if index_exists?(:taggings, :tag_id)
    remove_index :taggings, [:taggable_id, :taggable_type, :context]
    add_index :taggings,
              [:tag_id, :taggable_id, :taggable_type, :context, :tagger_id, :tagger_type],
              unique: true, name: 'taggings_idx'
  end

  def self.down
    remove_index :tags, :name

    remove_index :taggings, name: 'taggings_idx'

    add_index :taggings, :tag_id unless index_exists?(:taggings, :tag_id)
    add_index :taggings, [:taggable_id, :taggable_type, :context]
  end
```
This file is actually useful, I think I will add
```
    add_index :tags, :name, unique: true
    add_index :taggings,
              [:tag_id, :taggable_id, :taggable_type, :context, :tagger_id, :tagger_type],
              unique: true, name: 'taggings_idx'
```
to the migration file I have

Basically I think it's an overkill to have 5 migration files just for this 1 task: add tags

bboe: I really like that you questioned if the migrations were necessary. I like that you removed the ones pertaining to just mysql, however, I think you underestimate the importance of indexes. Even if you aren't taking advantaged of a certain column right now, the gem itself depends on all of its tables, and has those indexes for a reason so to prevent your app from potentially breaking in the future, it's a good idea to keep those pieces.

With that said, I think collapsing those multiple migrations into one is an awesome idea.

shfung: Really awesome that you have thought about the content of each of the 5 migrations and take the effort to combine them.  Just two comments:
> 4.add_missing_indexes

I don't think it's not very necessary.  Adding indices can be helpful to speed up queries.

> add_missing_unique_indices

I think we should also do this line `remove_index :taggings, :tag_id if index_exists?(:taggings, :tag_id)` as well

lsq921224:
@shfung Thanks for the comments. I added back indices to the db migration as well as a controller test and a unit test for adding tags. I also added a flow test to for the disappearing comma thing.

app/views/images/index.html.erb
```diff
@@ -2,4 +2,11 @@
 <% @images.each do |image| %>
   <%= image_tag(image.link, class: 'img_medium_display') %>
   <%= link_to :delete, image, method: :delete, data: { confirm: 'Are you sure?' } %>
+  <% if image.tag_list.present? %>
+    <div>
+    <% image.tags.each do |tag| %>
```
shfung: Do we want to show the tags on the `show` page as well?

test/controllers/images_controller_test.rb
shfung: This comment is not addressed yet.

lsq921224: Well I added a separate controller test for this.

test/models/image_test.rb
```diff
@@ -7,6 +7,11 @@ class ImageTest < ActiveSupport::TestCase
     assert_equal "can't be blank", image.errors.messages[:link].first
   end
 
+  test 'can save valid images with tags' do
+    image = Image.new link: 'http://google.com/a.jpeg', tag_list: 'tag1,tag2'
+    assert_predicate image, :valid?
```
shfung: Maybe assert the content of the tag list?

test/models/image_test.rb
```diff
@@ -7,6 +7,11 @@ class ImageTest < ActiveSupport::TestCase
     assert_equal "can't be blank", image.errors.messages[:link].first
   end
 
+  test 'can save valid images with tags' do
```
shfung: The test name is not exactly correctly because we haven't saved the image.

app/views/images/index.html.erb
lsq921224: I don't know, since it's not listed in the acceptance criteria, I did not implement that

test/controllers/images_controller_test.rb
shfung: For the controller tests, `tag_list` should be one of key of the params when calling to the controller.

lsq921224: well , i have this test
```
test 'should be able to create a valid image with tags' do
    assert_difference 'Image.count' do
      post images_path, params: { image: { link: 'http://www.w3schools.com/css/trolltunga.jpg',
                                           tag_list: 'tag1,tag2' } }
    end
    assert_equal 'Image created successfully.', flash[:notice]
    assert_redirected_to images_path
  end
```
Does it do the job?

shfung: This test is correct. 
But for the other tests,  `tag_list` should be one of key of the params when calling to the controller to better reflect the actual scenario. Because `tag_list` is now inside your form, so even if you don't input anything in the text field, the param that the `create` action received will still contain the key `tag_list` with an empty value.

lsq921224: Oh, Ok, so may I ask what would be the benefit to explicitly add it the the params?

shfung: So that the tests better reflect the actual behavior of the app.

test/controllers/images_controller_test.rb
```diff
@@ -5,14 +5,25 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  test 'should be able to create a valid image with tags' do
+    assert_difference 'Image.count' do
+      post images_path, params: { image: { link: 'http://www.w3schools.com/css/trolltunga.jpg',
+                                           tag_list: 'tag1,tag2' } }
+    end
+    assert_equal 'Image created successfully.', flash[:notice]
+    assert_redirected_to images_path
   end
 
   test 'should not be able to create an invalid image' do
```
shfung: Can you add `tag_list` as one of the params in this test as well?

lsq921224: Yes, I added it to all of them. XD

app/views/images/index.html.erb
rmacklin: Hmm, I don't know why the acceptance criterion specifically mentions _"on the index page."_ I'll remove that from the acceptance criterion. It is desired for tags to be displayed on the show page as well. More generally, you should be reusing a shared partial for the show and index pages, which will automatically ensure tags are displayed on both.

lsq921224: ok, since my "Add style" hasn't been merged yet, I will just add this change to that PR.

bboe: Please make the change its own pull request. Piggybacking work on one story to another adds complexity to the other story that shouldn't be there. Making a separate PR is a pretty cheap thing to do (assuming CI doesn't take hours).

app/controllers/images_controller.rb
```diff
@@ -10,7 +10,7 @@ def new
   def create
     @image = Image.new(image_params)
     if @image.save
-      redirect_to @image, notice: 'Image created successfully.'
+      redirect_to images_path, notice: 'Image created successfully.'
```
rmacklin: @lsq921224 Why was this changed? This behavior change is not part of "Display tags with associated images on the index page".

lsq921224: @rmacklin when I was working on this issue, the acceptance criterion  specifically mentioned something like after an image is created, show user the image on the index page.

rmacklin: Are you referring to this: https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/30#discussion_r73760699. Nothing about that old acceptance criteria said you needed to change where this redirects to.

rmacklin: Are you referring to this: https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/30#discussion_r73760699? Nothing about that old acceptance criteria said you needed to change where this redirects to. Furthermore, it's particularly bad to make a behavior change like that mixed in with lots of other changes in the "Display tags with associated images on the index page" commit. Such changes should be explicitly made in their own independent commits (though in this case, the change shouldn't have been made at all).

lsq921224: ok, I guess I misunderstood the acceptance criteria. Lessons learned!

rmacklin: The reason I am commenting on this now is because it's related to your question: https://github.com/AppFolioOnboarding/image-sharer-lsq921224/issues/33#issuecomment-241862094. As mentioned in my answer (https://github.com/AppFolioOnboarding/image-sharer-lsq921224/issues/33#issuecomment-241870346), you would not have run into that issue if you had not made this change.

rmacklin: Can you open a PR to undo this change? We want to follow the convention where a successful `create` redirects to `show`

lsq921224: I actually liked the fact I run into that issue. yes, I will open a PR for that.

PR #31, opened 2016-08-08
Display image tags on show page as well

lsq921224:
Display image tags on show page.


bboe:
Nearly perfect:

image

That appfolio tag belongs with the appfolio image. I know you haven't merged with your layout story, but even in this simple layout I would consider that a bug.

Also, can you replace the risque image from your seeds file (if that's where it originates from) and subsequently remove it from your stage and production servers?

lsq921224:
Thanks, i changed the layout and replaced the photo.

bboe:
👍

PR #32, opened 2016-08-15
Authentication

### LINKED ISSUE #11: Authentication
#### As an application hoster I want to add user authentication to my application.

- [ ] Users can sign up.
- [ ] Users can log in after they've signed up.
- [ ] Users can log out after logging in.
- [ ] Users can choose to be remembered after logging in.<sup>†</sup>
- [ ] Multiple browsers can concurrently hold permanent cookies for the same
  user.
- [ ] Logged in users should not be able to register, nor log in again.

<sup>†</sup>: "Remembering" means that users will remain logged into the site
even after completely closing their web browser.

__Note__: Do not use Devise for this task.

__Discussion topics__:
- [ ] If we wanted to add more advanced features such as email activation and
  "forgot my password", should we continue to "roll our own" authentication?

__Reference material__:
- https://www.railstutorial.org/book/log_in_log_out

__Further reading__:
- https://github.com/plataformatec/devise

__Dependencies__:
- Delete Image

lsq921224:
Closes #11


db/migrate/20160808181605_create_users.rb
```diff
@@ -0,0 +1,12 @@
+class CreateUsers < ActiveRecord::Migration[5.0]
+  def change
+    create_table :users do |t|
+      t.string :name, null: false
+      t.string :email, null: false
+      t.string :password_digest, null: false
```
bboe: ⭐ good job specifying `null: false`

test/models/user_test.rb
```diff
@@ -0,0 +1,74 @@
...
+                     password: 'foobar', password_confirmation: 'foobar')
+  end
+
+  test 'valid username and email' do
+    assert_predicate @user, :valid?
+  end
+
+  test 'username should be present' do
+    @user.name = ' '
+    assert_predicate @user, :invalid?
```
bboe: What else can you assert in this, and all other invalid tests?

test/models/user_test.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  test 'email should not be too long' do
+    @user.email = 'a' * 244 + '@example.com'
+    assert_predicate @user, :invalid?
+  end
+
+  test 'email validation should accept valid addresses' do
+    valid_addresses = %w(user@example.com USER@foo.COM A_US-ER@foo.bar.org
```
bboe: Does this list represent _all_ of the categories of valid addresses?

test/models/user_test.rb
```diff
@@ -0,0 +1,74 @@
...
+                           foo@bar_baz.com foo@bar+baz.com foo@bar..com)
+    invalid_addresses.each do |invalid_address|
+      @user.email = invalid_address
+      assert_predicate @user, :invalid?
+    end
+  end
+
+  test 'email addresses should be unique' do
+    duplicate_user = @user.dup
+    duplicate_user.email = @user.email.upcase
```
bboe: ⭐  Great test!

app/models/user.rb
```diff
@@ -0,0 +1,17 @@
+class User < ApplicationRecord
+  VALID_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z]+\z/i
+
+  before_save { email.downcase! }
+  validates :name,
+            presence: true,
+            length:   { maximum: 50 }
+  validates :email,
+            presence:   true,
+            length:     { maximum: 255 },
```
bboe: Is this maximum tested?

app/models/user.rb
```diff
@@ -0,0 +1,17 @@
+class User < ApplicationRecord
+  VALID_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z]+\z/i
+
+  before_save { email.downcase! }
+  validates :name,
+            presence: true,
+            length:   { maximum: 50 }
```
bboe: Is this maximum tested?

config/routes.rb
```diff
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   resources :images, only: [:show, :new, :create, :index, :destroy]
+  resources :users, only: [:new, :create]
```
bboe: Awesome.

app/models/user.rb
lsq921224: I actually have one test for this 
```
+  test 'name should not be too long' do
+    @user.name = 'a' * 51
+    assert_predicate @user, :invalid?
+  end
```
should I add more?

app/controllers/sessions_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+      flash.now[:session_error_msg] = 'Invalid email/password.'
+      render 'new', status: :unprocessable_entity
+    end
+  end
+  
+  private
+
+  def log_me_in(user)
+    log_in user
+    redirect_to images_path, notice: 'Log in successfully.'
```
bboe: `Logged`

app/models/user.rb
lsq921224: I actually have one test for this 
```
+  test 'name should not be too long' do
+    @user.name = 'a' * 51
+    assert_predicate @user, :invalid?
+  end
```
should I add more?

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  # Returns the current logged-in user (if any).
+  def current_user
+    @current_user ||= User.find_by(id: session[:user_id])
+  end
+
+  # Returns true if the user is logged in, false otherwise.
+  def logged_in?
+    !current_user.nil?
```
bboe: How you can rewrite this to avoid using `!` with a negative `predicate?` function?

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    get new_session_path
+    assert_response :success
+    assert_select 'form[action=?]', sessions_path
+  end
+
+  test 'login with invalid information' do
+    post sessions_path, params: { session: { email: '', password: '' } }
+    assert_response :unprocessable_entity
+    assert_select '.session_error_msg', 'Invalid email/password.'
+    assert_not logged_in?, 'User should not be logged in.'
```
bboe: Can you use `assert_predicate` here?

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+    user = User.create!(name:                  'foo',
+                        email:                 'foo@bar.com',
+                        password:              'password',
+                        password_confirmation: 'password')
+    post sessions_path, params: { session: { email:    user.email,
+                                             password: 'password' } }
+    assert_equal 'Log in successfully.', flash[:notice]
+    assert_redirected_to images_path
+    assert logged_in?, 'User should be logged in.'
+    assert_equal user.id, session[:user_id]
```
bboe: Do we need to specifically check the session?

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+  test 'login with valid information' do
+    user = User.create!(name:                  'foo',
+                        email:                 'foo@bar.com',
+                        password:              'password',
+                        password_confirmation: 'password')
+    post sessions_path, params: { session: { email:    user.email,
+                                             password: 'password' } }
+    assert_equal 'Log in successfully.', flash[:notice]
+    assert_redirected_to images_path
+    assert logged_in?, 'User should be logged in.'
```
bboe: Can we use `assert_predicate` here?

test/test_helper.rb
```diff
@@ -8,5 +8,9 @@ class TestCase
     fixtures :all
 
     # Add more helper methods to be used by all tests here...
+    # Returns true if a test user is logged in.
+    def logged_in?
+      !session[:user_id].nil?
```
bboe: Can we use a positive statement here?

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,29 @@
+require 'test_helper'
+
+class SessionsControllerTest < ActionDispatch::IntegrationTest
+  test 'should get login page' do
```
bboe: What if the user is already logged in?

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  test 'login with invalid information' do
+    post sessions_path, params: { session: { email: '', password: '' } }
+    assert_response :unprocessable_entity
+    assert_select '.session_error_msg', 'Invalid email/password.'
+    assert_not logged_in?, 'User should not be logged in.'
+  end
+
+  test 'login with valid information' do
```
bboe: What if the user is already logged in?

app/helpers/sessions_helper.rb
lsq921224: I thought about this. At first I wanted to use `present?`, but the behavior is not exactly the same. Maybe I can just do
```
def logged_in?
   current_user
end
```

app/controllers/sessions_controller.rb
```diff
@@ -11,7 +11,12 @@ def create
       render 'new', status: :unprocessable_entity
     end
   end
-  
+
+  def destroy
+    log_out
```
bboe: What if you're already logged out?

test/flows/signup_login_logout_test.rb
```diff
@@ -0,0 +1,34 @@
+require 'flow_test_helper'
+
+class SignupLoginLogoutTest < FlowTestCase
```
bboe: Can you add a little whitespace to break up the individual pages that are interacted with? It's a little hard to read through.

bboe: ⭐  great test otherwise

test/models/user_test.rb
```diff
@@ -71,4 +71,8 @@ def setup
     @user.password = @user.password_confirmation = 'a' * 5
     assert_predicate @user, :invalid?
   end
+
+  test 'authenticated? should return false for a user with nil digest' do
+    assert_not @user.authenticated?('')
```
bboe: Is there another assertion you can use which is not `assert_not`?

test/controllers/sessions_controller_test.rb
lsq921224: Yes, but then I will need to do something like
`    assert_predicate self, :logged_in?
`
and I find it harder to read than
`assert logged_in?, 'User should be logged in.'`

bboe:
Really great atomic commits! I added my first round of comments.

Additionally, I don't believe this acceptance test is met:

image

Try logging in permanently on two separate browsers, close them completely, reopen, and see if you're still logged in on both.

app/models/user.rb
bboe: Sorry, I missed that.

app/models/user.rb
bboe: Good, work. Sorry I missed that.

app/helpers/sessions_helper.rb
bboe: Is there anything about `present?` that doesn't work for what you want?

test/controllers/sessions_controller_test.rb
bboe: 👍  Great point. I like it the way you have it too.

test/models/user_test.rb
lsq921224: I'm not sure about this one. I can't use `assert_not_predicate` because `authenticated?` takes an argument.  `refute` will work but it does not really make a difference.

test/controllers/sessions_controller_test.rb
```diff
@@ -1,12 +1,18 @@
 require 'test_helper'
 
 class SessionsControllerTest < ActionDispatch::IntegrationTest
-  test 'should get login page' do
+  test 'should get login page if user is not logged in' do
     get new_session_path
     assert_response :success
```
bboe: Can you update this to `:ok`?

test/controllers/sessions_controller_test.rb
```diff
@@ -1,12 +1,18 @@
...
     assert_select 'form[action=?]', sessions_path
   end
 
+  test 'should not get login page if user is already logged in' do
+    SessionsController.any_instance.stubs(:logged_in?).returns(true)
```
bboe: ⭐  great use of stubbing.

test/controllers/sessions_controller_test.rb
```diff
@@ -1,12 +1,18 @@
...
+    get new_session_path
+    assert_redirected_to root_path
+  end
```
bboe: Is there a flash message to assert?

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,36 @@
...
+    assert_redirected_to images_path
+    assert logged_in?, 'User should be logged in.'
+  end
+
+  test 'login with valid information with remembering' do
+    user = create_a_user
+    log_in_as(user, remember_me: '1')
+    assert_not_nil cookies['user_id'], 'Cookie should not be nil'
+  end
+end
```
bboe: Are there any logout tests?

test/controllers/users_controller_test.rb
```diff
@@ -30,4 +30,15 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  test 'should not be able to create user if user is already logged in' do
+    UsersController.any_instance.stubs(:logged_in?).returns(true)
+    assert_no_difference 'User.count' do
+      post users_path, params: { user: { name:                  'user',
+                                         email:                 'user@valid.com',
+                                         password:              'foobar',
+                                         password_confirmation: 'foobar' } }
+    end
+    assert_redirected_to root_path
+  end
```
bboe: What about the new page when already logged in?

test/controllers/users_controller_test.rb
```diff
@@ -30,4 +30,15 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+
+  test 'should not be able to create user if user is already logged in' do
+    UsersController.any_instance.stubs(:logged_in?).returns(true)
+    assert_no_difference 'User.count' do
+      post users_path, params: { user: { name:                  'user',
+                                         email:                 'user@valid.com',
+                                         password:              'foobar',
+                                         password_confirmation: 'foobar' } }
+    end
+    assert_redirected_to root_path
```
bboe: Is there a message to assert?

app/helpers/sessions_helper.rb
```diff
@@ -4,9 +4,26 @@ def log_in(user)
...
+  end
+
   # Returns the current logged-in user (if any).
   def current_user
-    @current_user ||= User.find_by(id: session[:user_id])
+    if (user_id = session[:user_id])
+      @current_user ||= User.find_by(id: user_id)
+    elsif (user_id = cookies.signed[:user_id])
+      user = User.find_by(id: user_id)
+      if user.present?
```
bboe: Why is this check only done if the user_id is obtained through a signed cookie?

test/controllers/sessions_controller_test.rb
```diff
@@ -14,16 +14,17 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
...
     assert_equal 'Log in successfully.', flash[:notice]
     assert_redirected_to images_path
     assert logged_in?, 'User should be logged in.'
-    assert_equal user.id, session[:user_id]
+  end
+
+  test 'login with valid information with remembering' do
+    user = create_a_user
+    log_in_as(user, remember_me: '1')
+    assert_not_nil cookies['user_id'], 'Cookie should not be nil'
```
bboe: Should you do `assert_nil` in the without remembering test?

bboe: Should you `assert logged_in?`?

test/flows/signup_login_logout_test.rb
```diff
@@ -44,4 +48,38 @@ class SignupLoginLogoutTest < FlowTestCase
...
+    user = create_a_user
+    visit root_path
+    find("a[href='#{new_session_path}']").click
+    fill_in('session[email]', with: user.email)
+    fill_in('session[password]', with: user.password)
+    find(:css, '#session_remember_me').set(false)
+    find('input[type="submit"]').click
+    assert has_current_path?(images_path), "Does not have #{images_path}"
+    assert find('.notice').has_text?('Log in successfully.'), 'There should be a success message.'
+    page.driver.browser.manage.add_cookie(name: '_base_session', value: '')
```
bboe: ⭐  very clever

app/helpers/sessions_helper.rb
lsq921224: I copied and pasted this code from the book, I will make this logic clearer

bboe:
Overall this pull request looks awesome. The commits are very clean, and the pull request itself is easy to read through.

I added some inline feedback, and have one additional comment:

image

The above commit, while somehow became necessary to pass this story, should be in its own separate pull request. You don't need me to review that one, can merge it in, and then rebase this branch. The reason is that other stories will likely depend on that change, and if for whatever reason we need to revert this pull request after merging, we don't want to remove other things that are necessary.

lsq921224:
@bboe Thanks! I have addressed your comments!

test/controllers/sessions_controller_test.rb
```diff
@@ -0,0 +1,41 @@
...
+  test 'login with valid information without remembering' do
+    user = create_a_user
+    log_in_as(user, remember_me: '0')
+    assert_redirected_to images_path
+    assert_equal 'Log in successfully.', flash[:notice], 'There should be a notice'
+    assert logged_in?, 'User should be logged in.'
+    assert_nil cookies['user_id'], 'Cookie should not be nil'
+  end
+
+  test 'login with valid information with remembering' do
```
bboe: What about the logout tests?

test/controllers/users_controller_test.rb
```diff
@@ -0,0 +1,52 @@
+require 'test_helper'
+
+class ImagesControllerTest < ActionDispatch::IntegrationTest
+  test 'should get new user page' do
+    get new_user_path
+    assert_response :success
```
bboe: Please replace with `:ok` and look for other instances.

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def forget
+    cookies.delete(:user_id)
+  end
+
+  # Returns the current logged-in user (if any).
+  def current_user
+    user_id = session[:user_id]
+    user_id ||= cookies.signed[:user_id]
```
bboe: This line and the one above it can be combined into `user_id = session[:user_id] || cookies.signed[:user_id]`.

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+    user = User.find_by(id: user_id)
```
bboe: What if `user_id` is nil? Should we hit the database in such cases?

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+    return if user.nil?
+    log_in user
+    @current_user = user
+  end
+
+  # Returns true if the user is logged in, false otherwise.
+  def logged_in?
+    current_user.present?
```
bboe: Should this be `@current_user` so it doesn't require accessing the database?

lsq921224: I tried to change this to `@current_user`, but it does not work well because it seems I will lose `@current_user` when the page redirects; as a result, when I sign up a user and redirect to the home page, `@current_user` will give me nil even if I have set it to the new user.

lsq921224:
@bboe I addressed 4 of your 5 comments. I did not change `current_user` to `@current_user` because of the reason in my previous comment. I hope we can discuss this offline.

app/helpers/sessions_helper.rb
bboe: The typical approach here is to do something like the following in your function:

```
def current_user
  @current_user ||= find_current_user
end
```

That way your result is cached in the instance variable. The actual logic is only used at most once per request. Note that you don't need to create a private `find_current_user` function, but I think that's slightly cleaner than protecting the actual logic in a conditional.

app/helpers/sessions_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def forget
+    cookies.delete(:user_id)
+  end
+
+  # Returns the current logged-in user (if any).
+  def current_user
+    user_id = session[:user_id] || cookies.signed[:user_id]
+    return if user_id.nil?
```
bboe: You could remove this line (not important) if you rewrite the following line as:

    user = user_id && User.find_by(id: user_id)

In that case when `user_id` is nil it will return `nil` and when `user_id` is not nil it will return the result of the second part of the statement.

bboe:
Just the one comment about caching the current_user result. Looks great otherwise!

bboe:
👍

app/controllers/application_controller.rb
```diff
@@ -1,3 +1,4 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
+  include SessionsHelper
```
rmacklin: `SessionHelper` is intended to be used as the view helper for the `SessionController`'s views, and it's actually an antipattern to include that helper here in `ApplicationController`. I realize that this was actually suggested in the Rails Tutorial book, but I wanted to mention it.

A way you could work around this is to extract the methods that you need access to outside of the `SessionController` views into another module, and include that module where you need it as well as in `SessionHelper`. A benefit of that is that then you'll _only_ have access to the methods you actually want available exactly where you need them, rather than having access to all the methods in `SessionHelper` in every controller that inherits from `ApplicationController`.

bboe: I missed this before as well until Ryan was asking about how to test the helper. He extracted this behavior into a subclass of ActiveSupport::Concern.

Just to be clear, I don't have any expectations at this point that this code is updated, unless it's necessary for a future story.

lsq921224: I think almost everyone follows the Rails guide will make the same mistake, and I think something could be included in the story telling us that `helpers` are supposed to be view helpers.

I will fix this some time this week.

PR #34, opened 2016-08-16
Images by tag

### LINKED ISSUE #9: Images by Tag
#### As a user I want to view all images associated with a tag.

__Acceptance criteria__:
- [ ] When I click on a tag, I see a filtered list of only the images that have
  the tag I clicked on.
- [ ] The tag name (not its ID) is used to perform the filtering.

__Discussion topics__:
- [ ] Where should we implement this functionality?
- [ ] What happens if a user tries to manually filter by a nonexistent tag?

__Dependencies__:
- Image Tags

lsq921224:
Closes #9


shiboying:
Could you think another way to show images just on image#index page rather than create a new controller? So, you could think the tag is just like a filter. When you view index page, the image#index page basically show all the images. And imaging you hit a tag, what we want is basically show fewer images with this tag. Still like a index page right? So could you change to that way? 
Hint:`http://localhost:3000/images?tag=abc`
More discussion on this: https://github.com/AppFolioOnboarding/Boying_ImageUploader/pull/9#discussion_r51065358

lsq921224:
Thanks, this is a great point! I have updated the PR.

test/controllers/images_controller_test.rb
```diff
@@ -95,4 +95,21 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  test 'should show a filtered list of only the images that have the tag' do
+    image1 = Image.create!(link: 'http://www.w3schools.com/css/trolltunga.jpg', tag_list: 'tag,1')
+    image2 = Image.create!(link: 'http://www.w3schools.com/css/a.jpg', tag_list: 'tag,2')
+    image3 = Image.create!(link: 'http://www.google.com/a.jpg', tag_list: '3')
+    get images_path, params: { tag: 'tag' }
+    assert_response :success
+    assert_select 'img[src=?]', image1.link
+    assert_select 'img[src=?]', image2.link
+    assert_select 'img[src=?]', image3.link, count: 0
+  end
```
shiboying: It is better to separate setting test data and assertion with an empty line in tests.

And also, you could add a test case for `if the query contains 2 tags, always query by the first one`.

shiboying:
Do you already have flow tests? If you have, it would be better to add test cases for flow test.

test/controllers/images_controller_test.rb
lsq921224: By`if the query contains 2 tags, always query by the first one.`
Are you referring to cases like this: `http://localhost:3000/images?tag=tag&tag=tag2` ?

shiboying: Something like if you pass `params: {tag: 'tag1, tag2'}` it should check with `tag1`.

lsq921224: yeah, I can write a test case for that, it's already working that way though. For cases like this`http://localhost:3000/images?tag=tag1&tag=tag2`, the second param will override the first one (it will check tag2, not tag1), is this the expected behavior?

lsq921224: Oh, I see, right now params: {tag: 'tag1, tag2'} will also check with tag2

lsq921224:
Thanks Boying, I have made updates regarding your comments.

shiboying:
👍

app/controllers/images_controller.rb
```diff
@@ -1,6 +1,11 @@
 class ImagesController < ApplicationController
   def index
-    @images = Image.all
+    if params[:tag]
+      @images           = Image.tagged_with(params[:tag].split(',').first)
```
bboe: This indentation is very strange. Can you undo that? If it was done automatically, can you disable that feature?

lsq921224: Yes. I unchecked 'Align right parts of assignments or hashes'

PR #38, opened 2016-08-24
Redirect user to show page after creating an image

lsq921224:
Fix create action in images so that it redirects user to show page after creating an image instead of the index page


test/flows/images_crud_test.rb
```diff
@@ -18,13 +18,13 @@ class ImagesCrudTest < FlowTestCase
            'No error message: Link must start with either http or https and end with image extension'
     fill_in('image[link]', with: link)
     find('input[type="submit"]').click
-    assert has_current_path?(images_path), "Does not have #{images_path}"
+    assert has_current_path?(image_path(Image.last)), "Does not have #{image_path(Image.last)}"
```
rmacklin: The error message should say `Does not have path: #{image_path(Image.last)}`

rmacklin:
Looks fine besides the comment about the assertion failure message

PR #39, opened 2016-08-24
Add link to images in the index page

lsq921224:
Add links to images on the front page, so that a use is redirected to the show page when he clicks on the image


test/flows/images_crud_test.rb
```diff
@@ -91,11 +93,11 @@ class ImagesCrudTest < FlowTestCase
     fill_in('image[link]', with: link)
     find('input[type="submit"]').click
     visit(images_path)
-    find("a[href='/images/#{Image.last.id}']").click
+    find(".card-block > a[href=\"#{image_path(Image.last)}\"]").click
```
rmacklin: It would be clearer to the reader if you made this selector more specific by adding the data-method="delete" attribute to the selector. Then it becomes immediately clear that you're clicking on the delete button.

PR #40, opened 2016-08-25
Refactor flow tests to use page objects

### LINKED ISSUE #33: Refactor flow tests to use page objects
Reference material:
 - https://github.com/appfolio/ae_page_objects
 - https://sites.google.com/a/appfolio.com/eng/new-hire-resources/engineering-academy/ropes/image-sharing-project/flow-tests-with-page-objects

The starting skeleton is on the `flowTestsWithPageObjects` branch.

lsq921224:
Closes #33


app/views/images/_image.html.erb
```diff
@@ -1,6 +1,6 @@
-<div class="col-md-4">
+<div class="image-card-container col-md-4">
```
rmacklin: If this class is not used for styling, you should prefix it with `js-`

rmacklin: The same goes for all the classes you added here.

app/views/images/index.html.erb
```diff
@@ -1,3 +1,3 @@
-<div class="row">
+<div id="images-list" class="row">
```
rmacklin: `id`s do not need a `js-` prefix (because ids should never be used for styling), but our convention it to use underscores (not hyphens) as the word separator for ids.

test/flows/images_crud_test.rb
```diff
@@ -7,34 +7,35 @@ class ImagesCrudTest < FlowTestCase
     new_image_page = images_index_page.add_new_image!
 
     tags = %w(foo bar)
-    new_image_page = new_image_page.create_image!(url: 'invalid', tags: tags.join(', ')).as_a(PageObjects::Images::NewPage)
-    assert_equal 'must be a valid URL', new_image_page.url.error_message
+    new_image_page = new_image_page.create_image!(link: 'invalid',
+                                                  tags: tags.join(', ')).as_a(PageObjects::Images::NewPage)
```
rmacklin: If you're going to shorten this line, here's the recommended way:
```ruby
new_image_page = new_image_page.create_image!(
  link: 'invalid',
  tags: tags.join(', ')
).as_a(PageObjects::Images::NewPage)
```

Your current way is still quite long and has unnecessary whitespace

test/page_objects/images/show_page.rb
```diff
@@ -3,28 +3,31 @@ module Images
     class ShowPage < PageObjects::Document
       path :image
 
-      def has_url?(url)
-        #TODO
+      collection :tag_elements, locator: '.card-block', item_locator: '.image-tag'
```
rmacklin: Are you using this collection?

test/page_objects/images/show_page.rb
```diff
@@ -3,28 +3,31 @@ module Images
...
-        #TODO
+        node.click_on('delete')
         yield node.driver.browser.switch_to.alert
       end
 
       def delete_and_confirm!
-        #TODO
+        node.click_on('delete')
         alert = node.driver.browser.switch_to.alert
         alert.accept
```
rmacklin: You can refactor this to `delete(&:accept)`

app/views/tags/_tag_with_count.html.erb
```diff
@@ -1,4 +1,4 @@
 <div class="image-tag">
-  <%= link_to tag.name, "#{images_path(tag: tag.name)}" %>
+  <%= link_to tag.name, "#{images_path(tag: tag.name)}", class: 'image-tag__link' %>
```
rmacklin: As before, if this is not used for styling it should have a `js-` prefix, but additionally it should _not_ follow BEM syntax

test/flows/tags_flow_test.rb
```diff
@@ -2,19 +2,21 @@
...
+                   { link: cute_puppy_url, tag_list: 'puppy, cute' },
+                   { link: cute_puppy_url, tag_list: 'puppy' },
+                   { link: ugly_cat_url, tag_list: 'cat' }])
 
-    assert has_current_path?(images_path, only_path: true), "Does not have #{images_path}"
-    assert page.has_css?('img', count: 1), 'Page should have one image'
-    assert_equal 'http://www.w3schools.com/css/3.jpg', find('img')[:src],
-                 'Page is not showing the correct image'
+    tags_index_page = PageObjects::Tags::IndexPage.visit
+    assert tags_index_page.showing_tags_in_sequence?(tags: %w(puppy cute cat), counts: %w(3 2 1))
```
rmacklin: The interface I'd expect this page object to provide is:
```ruby
assert_equal [['puppy', 3], ['cute', 2], ['cat', 1]], tags_index_page.tags_with_count
```

test/flows/tags_flow_test.rb
```diff
@@ -2,19 +2,21 @@
...
+
+    cat_tag = tags_index_page.tag_list.find do |tag|
+      tag.name == 'cat'
+    end
+
+    images_index_page = cat_tag.view!
+    assert images_index_page.showing_image?(link: ugly_cat_url, tags: %w(cat))
```
rmacklin: You should also assert that the excluded images are _not_ showing.

test/flows/signup_login_logout_test.rb
```diff
@@ -6,81 +6,84 @@ def setup
...
-    find("a[href='#{session_path(User.last)}']").click
-    assert has_current_path?(root_path), "Does not have #{root_path}"
-    assert find('.alert.alert-success').has_text?('Log out successfully.'), 'There should be a success message.'
+    images_index_page = PageObjects::Images::IndexPage.visit
+    new_user_page = images_index_page.sign_up!
+
+    new_user_page = new_user_page.create_user!(name: '',
+                                               email: '',
+                                               password: '',
+                                               password_confirmation: '').as_a(PageObjects::Users::NewPage)
```
rmacklin: Same as before:
```ruby
new_user_page = new_user_page.create_user!(
  name: '',
  email: '',
  password: '',
  password_confirmation: ''
).as_a(PageObjects::Users::NewPage)
```

test/flows/signup_login_logout_test.rb
```diff
@@ -6,81 +6,84 @@ def setup
...
+
+    images_index_page = new_user_page.create_user!(name: name,
+                                                   email: email,
+                                                   password: password,
+                                                   password_confirmation: password)
+    assert_equal 'Account created successfully.', images_index_page.flash_message(:success)
+
+    assert images_index_page.logout_link.present?
+    refute images_index_page.signup_link.present?
+    refute images_index_page.login_link.present?
```
rmacklin: You should use the predicate assertions to make the failure messages better:
```ruby
assert_predicate images_index_page.logout_link, :present?
refute_predicate images_index_page.signup_link, :present?
refute_predicate images_index_page.login_link, :present?
```
which is equivalent to:
```ruby
assert_predicate images_index_page.logout_link, :present?
assert_predicate images_index_page.signup_link, :absent?
assert_predicate images_index_page.login_link, :absent?
```
but `present?` doesn't check for visibility, so this should really be:
```ruby
assert_predicate images_index_page.logout_link, :visible?
assert_predicate images_index_page.signup_link, :absent?
assert_predicate images_index_page.login_link, :absent?
```

test/flows/signup_login_logout_test.rb
```diff
@@ -6,81 +6,84 @@ def setup
...
+
+    images_index_page = images_index_page.log_out!
+    assert_equal 'Log out successfully.', images_index_page.flash_message(:success)
+
+    refute images_index_page.logout_link.present?
+    assert images_index_page.signup_link.present?
+    assert images_index_page.login_link.present?
```
rmacklin: Same as above

test/flows/signup_login_logout_test.rb
```diff
@@ -6,81 +6,84 @@ def setup
...
-
-    find('input[type="submit"]').click
-    assert has_current_path?(images_path), "Does not have #{images_path}"
-    assert find('.alert.alert-success').has_text?('Log in successfully.'), 'There should be a success message.'
-
-    find("a[href='#{session_path(User.last)}']").click
-    assert has_current_path?(root_path), "Does not have #{root_path}"
-    assert find('.alert.alert-success').has_text?('Log out successfully.'), 'There should be a success message.'
+    images_index_page = PageObjects::Images::IndexPage.visit
+    new_user_page = images_index_page.sign_up!
```
rmacklin: Flow tests should use the terminology a user would think in. For example, a user doesn't think "new user page"; they think "sign up page", so our tests and page objects should use names aligned with that.

test/flows/signup_login_logout_test.rb
```diff
@@ -6,81 +6,84 @@ def setup
...
+    refute images_index_page.login_link.present?
+
+    images_index_page = images_index_page.log_out!
+    assert_equal 'Log out successfully.', images_index_page.flash_message(:success)
+
+    refute images_index_page.logout_link.present?
+    assert images_index_page.signup_link.present?
+    assert images_index_page.login_link.present?
+
+    new_session_page = images_index_page.log_in!
```
rmacklin: As above, this should be `login_page`, not `new_session_page`, to match with how users would think about interacting with our application in this flow.

test/flows/signup_login_logout_test.rb
```diff
@@ -6,81 +6,84 @@ def setup
...
-    assert has_current_path?(images_path), "Does not have #{images_path}"
-    assert find('.alert.alert-success').has_text?('Log in successfully.'), 'There should be a success message.'
-
-    find("a[href='#{session_path(User.last)}']").click
-    assert has_current_path?(root_path), "Does not have #{root_path}"
-    assert find('.alert.alert-success').has_text?('Log out successfully.'), 'There should be a success message.'
+    images_index_page = PageObjects::Images::IndexPage.visit
+    new_user_page = images_index_page.sign_up!
+
+    new_user_page = new_user_page.create_user!(name: '',
```
rmacklin: Similarly this should be `sign_up!`, not `create_user!`, which matches the text of the button that's being clicked on.

test/flows/signup_login_logout_test.rb
```diff
@@ -6,81 +6,84 @@ def setup
...
+
+    images_index_page = images_index_page.log_out!
+    assert_equal 'Log out successfully.', images_index_page.flash_message(:success)
+
+    refute images_index_page.logout_link.present?
+    assert images_index_page.signup_link.present?
+    assert images_index_page.login_link.present?
+
+    new_session_page = images_index_page.log_in!
+    new_session_page = new_session_page.create_session!(email: '',
```
rmacklin: Similarly, this should be `log_in!`, not `create_session!`, which matches the text of the button that's being clicked on.

test/page_objects/document.rb
```diff
@@ -1,7 +1,26 @@
...
+    element :signup_link, locator: '.signup-link'
+    element :login_link, locator: '.login-link'
+    element :logout_link, locator: '.logout-link'
+
     def flash_message(message_type)
       node.find(".alert.alert-#{message_type}").text
     end
+
+    def sign_up!
+      node.click_link('Sign up')
```
rmacklin: Why aren't you using the `sign_up_link` element here? Same for the other methods.

test/page_objects/sessions/new_page.rb
```diff
@@ -0,0 +1,23 @@
+module PageObjects
+  module Sessions
+    class NewPage < PageObjects::Document
```
rmacklin: Following the previous discussion, this should be `PageObjects::LoginPage`, rather than `PageObjects::Sessions::NewPage`, to match how users think about the application, not how it's implemented.

test/page_objects/users/new_page.rb
```diff
@@ -0,0 +1,31 @@
+module PageObjects
+  module Users
+    class NewPage < PageObjects::Document
```
rmacklin: Following the previous discussion, this should be `PageObjects::SignupPage`, rather than `PageObjects::Users::NewPage`, to match how users think about the application, not how it's implemented.

test/flows/signup_login_logout_test.rb
```diff
@@ -6,81 +6,84 @@ def setup
...
+                                                        password: '',
+                                                        remember_me: 'off').as_a(PageObjects::Sessions::NewPage)
```
rmacklin: No need to pass a `remember_me` option if it's not being set.

test/page_objects/sessions/new_page.rb
```diff
@@ -0,0 +1,23 @@
...
+      form_for :session do
+        element :email
+        element :password
+        element :remember_me
+      end
+
+      def create_session!(email: nil, password: nil, remember_me: nil)
+        self.email.set email if email.present?
+        self.password.set password if password.present?
+        self.remember_me.set true if remember_me == 'on'
```
rmacklin: Let's change the parameter to `remember_me: false` change the interface to just passing `true` when it should be checked, which would change this line to:
```ruby
self.remember_me.set true if remember_me
```

test/page_objects/users/new_page.rb
```diff
@@ -0,0 +1,31 @@
...
+        self.email.set email if email.present?
+        self.password.set password if password.present?
+        self.password_confirmation.set password_confirmation if password_confirmation.present?
+
+        click_sign_up!
+      end
+
+      private
+
+      def click_sign_up!
```
rmacklin: What is the purpose of this method?

test/flows/signup_login_logout_test.rb
lsq921224: In one of my previous PR, I got a comment about passing unused parameters  even if they are not being used, so that the test better reflects what the app is doing. https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/30#discussion_r73762460
Even though that comment was on a controller test, I do think it makes sense in flow tests as well, so I added` remember_me` here.

test/page_objects/users/new_page.rb
lsq921224: `rubocop` was complaining that `Assignment Branch Condition Size` is too high for `def create_user!(name: nil, email: nil, password: nil, password_confirmation: nil)`, I couldn't find a better way to refactor that method, so I abstracted these two lines of code

test/flows/signup_login_logout_test.rb
rmacklin: > Even though that comment was on a controller test, I do think it makes sense in flow tests as well

Respectfully, I completely disagree. In a controller test it makes _perfect sense_ to make your requests match what requests the actual application will make. But in a flow tests, you should match what a user would actually do (which in this case is: completely ignore the remember_me checkbox). The request the controller receives will still be the same.

test/page_objects/users/new_page.rb
rmacklin: I don't see how extracting these two non-conditional lines would fix a "too many conditions" error.

lsq921224: well, I think it helps reduce `branch`, http://c2.com/cgi/wiki?AbcMetric, as a result it fixes this error

test/flows/signup_login_logout_test.rb
lsq921224: I see

test/page_objects/users/new_page.rb
rmacklin: I have to say I just don't agree that this method's complexity is any better with the private method. I think it's worse this way. Rubocop is great, but I'm not a fan of making code less direct/readable just to silence its complaint. If you wish to keep this rule configured as is, I think a better abstraction would be to extract a method that DRYs the repeated pattern:
```ruby
self.<thing>.set <thing> if <thing>.present?
```

However, I honestly think the method is just fine without extracting anything.

bboe: @lsq921224 please add this file to your repo: https://github.com/AppFolioOnboarding/base/blob/master/test/.rubocop.yml

Some of the metrics don't make sense to have in the test code.

lsq921224:
@rmacklin I think I addressed your comments and I updated the PR.

test/page_objects/tags/index_page.rb
```diff
@@ -0,0 +1,18 @@
...
+
+      collection :tag_list, locator: '#tags-list', item_locator: '.image-tag', contains: ImageTag do
+        def view!
+          find('.js-image-tag-link').click
+          window.change_to(PageObjects::Images::IndexPage)
+        end
+      end
+
+      def tags_with_count
+        [tag_list.map(&:name), tag_list.map(&:count)].transpose
```
rmacklin: This is a pretty neat implementation. It does `map` over the same collection twice and then perform a transpose, though, so there is definitely going to be redundant iteration which is something you should always look to avoid if it doesn't sacrifice readability. In this case, you can iterate over the collection just once without losing readability using:

```ruby
tag_list.map { |tag| [tag.name, tag.count] }
```

test/page_objects/images/new_page.rb
```diff
@@ -0,0 +1,20 @@
...
+      form_for :image do
+        element :link
+        element :tag_list
+      end
+
+      def create_image!(link: nil, tags: nil)
+        self.link.set link if link.present?
+        tag_list.set tags if tags.present?
+        node.click_button('Create Image')
+        window.change_to(ShowPage, NewPage)
```
rmacklin: Typically this is done as
```ruby
window.change_to(ShowPage, self.class)
```
which indicates "this page is either going to change to the show page, or it'll rerender this page again". Additionally, if you were to rename `NewPage`, you wouldn't have to modify this line.

The same comment applies to the other page objects that have themselves as an argument to `window.change_to`

test/page_objects/tags/index_page.rb
lsq921224: This is great!

lsq921224:
@rmacklin I have updated the PR.

Issue #10, opened 2016-07-19
Share Image


AppFolioRopes:
#### As a user I want to share an image with someone via email.

__Story__:

This application is called Image Sharer. Sure, you can share images by adding
them to our web service, but how will people who do not know about this web
service be able to see those images? In phone calls with our customers we have
discovered that many of our users are manually emailing their friends links to
the images. That is awesome! We can make things more efficent for our users by
directly supporting sharing images through email.

__Acceptance criteria__:
- [x] There is a link to share an image.
- [x] After clicking on this link, a bootstrap modal appears with a form to
  send the email.
- [x] The form has fields for the recipient's email and an optional message.
- [x] The email we send has an HTML part and a text part.
- [x] The email contains the image, the custom message if provided, and a link
  inviting the recipient to check out our application.
- [x] The modal closes and a success message is shown if the email was sent
  successfully.
- [x] The user is alerted of errors if the operation was unsuccessful.

__Note__: This story is large, so it makes sense to break the implementation
down into more manageable pieces (a la elephant carpaccio). Please start by
implementing a basic sharing feature that uses a new page (not a modal), and
open a PR for that. Then, once we've verified the simplified version is
working, refactor it to use a modal, and open a second PR for that. The
[video](https://sites.google.com/a/appfolio.com/eng/new-hire-resources/engineering-academy/ropes/image-sharing-project/basics-of-bootstrap-modals-js-and-ajax)
provides some background and code examples that will be really helpful for the
second PR, so it's worth watching (and following along) before you begin that
part.

__Discussion topics__:
- [ ] How could we develop the front end and back end separately? (And why
  might that be a good idea?)
- [ ] How should we implement a non-CRUD action?
- [ ] What are the different types of errors we should handle?
- [ ] What is AJAX? What libraries can help us add AJAX to our Rails app?
- [ ] How should we go about sending emails from our Rails app?
- [ ] How could we show a preview of the image being shared in the modal?
- [ ] What are the implications of allowing users to send emails from our app?

__Reference material__:
- http://guides.rubyonrails.org/action_mailer_basics.html
- https://elements.heroku.com/addons/sendgrid
- https://devcenter.heroku.com/articles/sendgrid#ruby-rails
- https://github.com/rails/jquery-ujs/wiki
- http://v4-alpha.getbootstrap.com/components/modal
[...]

PR #43, opened 2016-08-29
Share image part1

lsq921224:
Part 1 of #10


app/models/share_mail_form.rb
```diff
@@ -0,0 +1,11 @@
+class ShareMailForm
```
rmacklin: This is a form object, not a model, so we shouldn't put it in app/models.

app/models/share_mail_form.rb
```diff
@@ -0,0 +1,11 @@
...
+  include ActiveModel::Model
+
+  VALID_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z]+\z/i
```
rmacklin: This is duplicated: https://github.com/AppFolioOnboarding/image-sharer-lsq921224/blob/6e36d3247c23b429d80b971cb8cf0758dc0debd1/app/models/user.rb#L2 . We should DRY it

app/assets/stylesheets/_image-card.scss
```diff
@@ -3,9 +3,13 @@
   border-radius: 5px;
   margin-bottom: 10px;
 }
+.image-card__buttons {
+  border-top: 1px solid #ececec;
+  padding-top: 10px;
+}
```
rmacklin: This, and other view-related changes, do not belong in an "Add share action for images" commit

app/views/images/_image.html.erb
```diff
@@ -9,7 +9,10 @@
           <%= render partial: 'tags/tag', locals: { tag: tag } %>
         <% end %>
       <% end %>
-      <%= link_to :delete, image, method: :delete, data: { confirm: 'Are you sure?' }, class: 'btn btn-outline-danger pull-xs-right' %>
+      <div class="image-card__buttons">
+        <%= link_to :share, share_image_path(image), class: 'btn btn-outline-primary' %>
```
rmacklin: This works, but I want to mention that it's not common to use symbols instead of strings as the first argument of `link_to`

app/views/images/share/form.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= simple_form_for @share_mail_form, url: share_image_path(@image) do |f| %>
+  <div class="form-group row">
+    <%= link_to image_path(@image), class: 'offset-sm-2' do %>
+      Image to share
+      <%= image_tag(@image.link, class: 'image-card__image') %>
```
rmacklin: This breaks BEM. Talk to @aaronpanch if you need a refresher.

app/views/images/share/form.html.erb
```diff
@@ -0,0 +1,13 @@
...
+    <% end %>
+  </div>
+  <%= f.input :email, required: true %>
```
rmacklin: Is it necessary to manually specify `required: true` for this field?

app/views/images/share/form.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <%= f.input :message %>
+  <div class="form-group row">
+    <%= f.button :submit, 'Share' %>
```
rmacklin: FYI, you can also use `f.submit 'Share'`

test/controllers/images_controller_test.rb
```diff
@@ -112,4 +112,20 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+
+    assert_response :ok
+    assert_select 'img[src=?]', image.link
+    assert_select '#new_share_mail_form'
+  end
+
+  test 'should not get share image page if the image does not exist' do
+    get share_image_path(-1)
+    assert_equal 'Image not found.', flash[:error]
+    assert_redirected_to images_path
```
rmacklin: Minor, but we usually prefer to assert about th response status prior to any other assertions.

app/mailers/image_mailer.rb
```diff
@@ -0,0 +1,10 @@
+class ImageMailer < ApplicationMailer
+  default from: 'sharing@imagesharer.com'
```
rmacklin: Minor, but I wouldn't use this domain that isn't actually yours

app/mailers/image_mailer.rb
```diff
@@ -0,0 +1,10 @@
...
+
+  def image_email(image, recipient, message)
```
rmacklin: How about `share_image_email`? And can you use keyword arguments?

app/mailers/image_mailer.rb
```diff
@@ -0,0 +1,10 @@
...
+    @image = image
+    @message = message
+
+    mail(to: recipient, subject: 'Checkout this image')
```
rmacklin: Minor, but "check out" should be two words here

app/views/image_mailer/image_email.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1>Checkout this awesome image!</h1>
+<%= image_tag(@image.link) %>
+<div>
+  <div style="font-size: 150%;">
+    <%= @message %>
```
rmacklin: What will happen if they didn't provide a message?

app/views/image_mailer/image_email.text.erb
```diff
@@ -0,0 +1,8 @@
+Checkout this awesome image
+===============================================
+
+Link: <%= @image.link %>
+
+<%= @message %>
```
rmacklin: What will happen if they didn't provide a message?

app/views/image_mailer/image_email.html.erb
```diff
@@ -0,0 +1,10 @@
...
+  </div>
+  <div style="color:red">
+    <%= link_to 'Image sharer', root_url %>
```
rmacklin: As of this commit, `root_url` won't work inside a mailer.

config/environments/development.rb
```diff
@@ -27,9 +27,16 @@
...
+  config.action_mailer.raise_delivery_errors = true
 
   config.action_mailer.perform_caching = false
+  config.action_mailer.delivery_method = :sendmail
+  config.action_mailer.smtp_settings = {
+    address: 'localhost',
+    port: 25,
+    domain: 'localhost:3000'
+  }
+  config.action_mailer.default_url_options = { host: 'localhost:3000' }
```
rmacklin: These config changes can be their own commit, prior to this one.

test/mailers/image_mailer_test.rb
```diff
@@ -0,0 +1,23 @@
...
+    assert_equal ['sharing@imagesharer.com'], email.from
+    assert_equal ['friend@example.com'], email.to
+    assert_equal 'Checkout this image', email.subject
+
+    [:text_part, :html_part].each do |part|
+      email_body = email.send(part).body.to_s
+      assert email_body.include?(image.link), 'Email body does not contain image link'
+      assert email_body.include?(optional_message), 'Email body does not contain optional message'
+    end
+  end
```
rmacklin: There should be a test for the case where there is no message provided

test/mailers/image_mailer_test.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    assert_equal ['sharing@imagesharer.com'], email.from
+    assert_equal ['friend@example.com'], email.to
+    assert_equal 'Checkout this image', email.subject
+
+    [:text_part, :html_part].each do |part|
+      email_body = email.send(part).body.to_s
+      assert email_body.include?(image.link), 'Email body does not contain image link'
+      assert email_body.include?(optional_message), 'Email body does not contain optional message'
```
rmacklin: We should also assert that the link to our application is included

app/controllers/images_controller.rb
```diff
@@ -46,9 +46,34 @@ def share
...
+
+  def deliver_mail(image, share_mail_form)
+    ImageMailer.image_email(image, share_mail_form.email, share_mail_form.message).deliver_now!
+  rescue StandardError
+    flash.now[:error] = 'Problems sending email.'
+    render 'images/share/form', status: :internal_server_error
+  else
+    flash.now[:success] = 'Message sent.'
+    render 'images/share/form'
+  end
```
rmacklin: A few things: First, `rescue/else` is not a common pattern, and I think it should be avoided. Second, I think we can just let unexpected errors bubble up and be handled by the default error handling, which will simplify this code to the point where you probably don't need this private method.

test/controllers/images_controller_test.rb
```diff
@@ -128,4 +128,42 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    end
+    share_email = ActionMailer::Base.deliveries.last
+
+    assert_equal 'Checkout this image', share_email.subject
+    assert_equal 'friend@example.com', share_email.to[0]
+
+    [:text_part, :html_part].each do |part|
+      email_body = share_email.send(part).body.to_s
+      assert email_body.include?(image.link), 'Email body does not contain image link'
+      assert email_body.include?('this is a test'), 'Email body does not contain optional message'
```
rmacklin: Here as well you can also include an assertion that the email correctly contains a link to our application

app/controllers/images_controller.rb
rmacklin: Third, currently there is no test coverage for the error handling.

config/application.rb
```diff
@@ -17,10 +17,11 @@ class Application < Rails::Application
...
     csp_settings = ["connect-src 'self'",
                     "default-src 'none'",
-                    'font-src https://cdn.jsdelivr.net',
+                    "font-src https://cdn.jsdelivr.net 'self'",
                     "img-src 'self' http: https:",
-                    "script-src 'self' 'unsafe-eval'",
-                    "style-src 'self' https://cdn.jsdelivr.net"]
+                    "script-src 'self' 'unsafe-eval' 'unsafe-inline'",
+                    "frame-src 'self'",
+                    "style-src 'self' https://cdn.jsdelivr.net 'unsafe-inline'"]
```
rmacklin: I'm guessing these changes were made to support the mailer preview. That's cool, but:

1. this commit should be before the commit that adds the mailer preview
2. did you run these changes by @bboe?

app.json
```diff
@@ -18,10 +18,17 @@
...
+      "required": true
+    },
+    "SENDGRID_PASSWORD": {
+      "required": true
     }
   },
   "addons": [
-    "heroku-postgresql"
+    "heroku-postgresql",
+    "sendgrid"
```
rmacklin: I think this should be `"sendgrid:starter"` to ensure we use the free "starter" plan. In fact, I think `"heroku-postgresql"` should actually be `"heroku-postgresql:hobby-dev"`. Maybe heroku applies the free plans as the default, but I prefer being explicit in case that changes.

rmacklin: Additionally, I would move this commit earlier in the history, prior to you referencing the SendGrid environment variables.

test/flows/share_image_flow_test.rb
```diff
@@ -0,0 +1,23 @@
...
+    image_show_page = image_to_share.view!
+
+    image_share_page = image_show_page.share!
+
+    image_share_page = image_share_page.share!
+    assert_equal "Email can't be blank", image_share_page.email.error_message
+
+    image_share_page = image_share_page.share!(email: 'friend@example.com', message: 'hello')
+    assert_equal 'Message sent.', image_share_page.flash_message(:success)
+  end
```
rmacklin: What other case should be tested?

test/flows/share_image_flow_test.rb
```diff
@@ -0,0 +1,23 @@
...
+    Image.create!(link: puppy_url, tag_list: 'superman, cute')
+
+    images_index_page = PageObjects::Images::IndexPage.visit
+    image_to_share = images_index_page.images.find do |image|
+      image.url == puppy_url
+    end
+
+    image_show_page = image_to_share.view!
+
+    image_share_page = image_show_page.share!
```
rmacklin: This is testing that images can be shared from the show page. They can also be shared from the index page.

app/controllers/images_controller.rb
```diff
@@ -46,9 +46,34 @@ def share
...
+
+  def deliver_mail(image, share_mail_form)
+    ImageMailer.image_email(image, share_mail_form.email, share_mail_form.message).deliver_now!
+  rescue StandardError
+    flash.now[:error] = 'Problems sending email.'
+    render 'images/share/form', status: :internal_server_error
+  else
+    flash.now[:success] = 'Message sent.'
+    render 'images/share/form'
```
rmacklin: I'm curious, why did you choose to remain on the share form after successfully sharing an image?

test/flows/share_image_flow_test.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    image_share_page = image_share_page.share!
+    assert_equal "Email can't be blank", image_share_page.email.error_message
+
+    image_share_page = image_share_page.share!(email: 'friend@example.com', message: 'hello')
```
rmacklin: It's a good idea to test that the form preserves user input when there is a validation error. So you could fill in an invalid email and a message, and then verify that the input was preserved and then (rather than calling `share!` with arguments) manually set just the email field and resubmit the form.

rmacklin:
Nice job; this is a fantastic start! Most of my comments are pretty minor. You did a pretty good job splitting out the commits which made the review process easier, but (as mentioned in my comments) I think there are still a few improvements we can make to the commit history.

config/environments/production.rb
```diff
@@ -56,6 +56,18 @@
...
+  config.action_mailer.smtp_settings = {
+    address:              'smtp.sendgrid.net',
+    port:                 '587',
+    authentication:       :plain,
+    user_name:            ENV['SENDGRID_USERNAME'],
+    password:             ENV['SENDGRID_PASSWORD'],
+    enable_starttls_auto: true
+  }
+
+  config.action_mailer.default_url_options = { host: ENV['MAILER_HOST'] }
```
rmacklin: I'm curious, where did you get the idea for this? It's the right idea but it looks like you haven't set it up because the link in the emails sent from your review app don't work

app/controllers/images_controller.rb
lsq921224: I did that because I thought the user might want to share the image to another friend after successfully sharing an image.

config/application.rb
bboe: > unsafe-inline

It would be great to avoid adding unsafe-inline, especially for script-src since that opens your application up to XSS vulnerabilities. Can you tell me why it's needed, and how much effort it would require to work around it?

config/environments/production.rb
lsq921224: I got this idea because I manually deployed this branch on Heroku on lsq921224s.herokuapp.com, then when I created a PR, I had a new host lsq921224-pr-43.herokuapp.com, so my previous setting (setting it to a particular host url) stopped making sense.

app/controllers/images_controller.rb
rmacklin: Do you think that is more likely than wanting to share a different image?

config/environments/production.rb
rmacklin: OK, but you haven't set it up properly because the email links aren't working.

lsq921224: I did set it up but it looks like it got lost. My guess is every time I push a update, I will lose the config vars I manually added on heroku. I may modify my app.json to make this work.

config/application.rb
lsq921224: It's needed for the mailer preview. Without 'unsafe-inline' for script-src, I lost the ability to choose between html-preview and text-preview. Without 'unsafe-inline' for style-src, I will get a very small text box for my email body
image

In the chrome error message, it gives me three options 

> Either the 'unsafe-inline' keyword, a hash ('sha256-yk37m0UlD5Y1tFbZ+3TY7UWMatMsVyKZYpXp7jfirAA='), or a nonce ('nonce-...') is required to enable inline execution.

'unsafe-inline' is the quickest solution, I am looking into the other two now since 'unsafe-inline' is not the recommended way.

bboe: As we discussed on slack. It seems that this preview is not being directly used in production, but only to help aid development. As a result I don't think that these CSP changes are warranted. Furthermore, I think it is important ensure that the mailer preview routes are not accessible in production mode since we do not intend for our users to access them.

app/assets/stylesheets/_image-card.scss
lsq921224: I change the commit message to "Implement share page for each image", did you want me to split this into separate commits or is this new commit message fine?

rmacklin: Since the commit can be split into two atomic commits, I would suggest doing that.

app/views/images/share/form.html.erb
lsq921224: Yeah, if I use f.button, my config for simple_form   

> config.button_class             = 'btn btn-default offset-sm-2'
will apply automatically to the button; however, if I use f.submit instead, I don't have that anymore.

app/views/image_mailer/image_email.html.erb
lsq921224: Should I add a default message or not show this div if there's no message?

config/application.rb
lsq921224: @bboe I did some research and it looks like the mailer preview is not accessible in production by default. 

Instead of changing the CSP setting in application.rb, I add these changes to config/development.rb which will override application.rb, so I can use the nice previewer in development without affecting production. Do you think this is ok?

bboe: @lsq921224 yes, I think that would be suitable.

app/views/images/share/form.html.erb
rmacklin: Is there no analogous config that can be set? Now that you mention it, submit buttons should have a "primary" button style, rather than a "default" button style.

app/views/image_mailer/image_email.html.erb
rmacklin: That's up to you. A default is probably not necessary since you have the "Check out this awesome image!" heading, but either way you'll have to implement conditional logic in the mailer (which is part of the exercise)

app/views/images/share/form.html.erb
lsq921224: unfortunately I couldn't find any analogous config; all I find is `error_notification_class`, `button_class`, `boolean_label_classs` and `default_wrapper`. 
Simple-form does have setting for `input` in general, for example, mine currently is
`    b.use :input, class: 'col-sm-4 form-control'
`
but this applies to all `input` tags

I didn't know "btn-default" is dropped in boostrap v4, I will change that to "btn-secondary" though since it fits the theme of my app better than "btn-primary"

rmacklin: It's better to use `btn-primary` since semantically this _is_ a primary button. You can configure bootstraps variables to change the appearance so it fits your theme.

lsq921224: Oh, thank you for pointing this out

lsq921224:
I think I have addressed the comments; the commit history is out of order though because of rebasing, and I don't know if there is a good way to show them in order

rmacklin:
`git rebase --ignore-date` should help with that. It's really annoying that GitHub PRs do not respect the commit tree order.

lib/constants/email.rb
```diff
@@ -0,0 +1,5 @@
+module Constants
+  module Email
+    VALID_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z]+\z/i
+  end
+end
```
rmacklin: When you're extracting something to share between two places, you should take some time to think about exactly what it is that should be extracted. In this case, you've extracted a `VALID_EMAIL_REGEX` constant, which helps some, but I think it would be more appropriate to extract a custom email validator (which would include the regex) and use that in both places that need to validate emails.

config/initializers/simple_form_bootstrap.rb
```diff
@@ -1,7 +1,7 @@
 # Use this setup block to configure all options available in SimpleForm.
 SimpleForm.setup do |config|
   config.error_notification_class = 'alert alert-danger'
-  config.button_class             = 'btn btn-default offset-sm-2'
+  config.button_class             = 'btn btn-primary offset-sm-2'
```
rmacklin: So, this isn't what I was suggesting earlier. What I said before was that _submit_ buttons should always be primary. This is going to make _all_ simple_form-generated buttons be primary. You should add the `btn-primary` class to your submit buttons.

app/views/image_mailer/share_image_email.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1>Checkout this awesome image!</h1>
```
rmacklin: I see that you changed the email subject to say "Check out" but the mailer template still says "Checkout"

app.json
```diff
@@ -21,7 +21,8 @@
     }
   },
   "addons": [
-    "heroku-postgresql"
+    "heroku-postgresql:hobby-dev",
+    "sendgrid:starter"
```
rmacklin: This may seem minor, but technically there are two distinct changes in this commit: specifying the addon level for the postgres addon, and adding the sendgrid addon. These are distinct enough that they can be their own commits, but at the very least, both should be mentioned in the commit message if they are grouped into a single commit.

test/flows/images_crud_test.rb
```diff
@@ -44,12 +44,12 @@ class ImagesCrudTest < FlowTestCase
...
     image_show_page = image_to_delete.view!
 
-    image_show_page.delete do |confirm_dialog|
+    image_show_page.image.delete do |confirm_dialog|
       assert_equal 'Are you sure?', confirm_dialog.text
       confirm_dialog.dismiss
     end
 
-    images_index_page = image_show_page.delete_and_confirm!
+    images_index_page = image_show_page.image.delete_and_confirm!
```
rmacklin: You can avoid these changes if you make use of the `delegate` method to
```ruby
delegate :delete, :delete_and_confirm!, to: :image
```
This keeps preserves the nice page object interface of your `ShowPage`, while still enabling you to DRY the delete methods by letting the `ImageCard` handle their implementation.

test/flows/share_image_flow_test.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    image_share_page = image_share_page.share!.as_a(PageObjects::Images::SharePage)
+    assert_equal "Email can't be blank", image_share_page.email.error_message
+
+    image_share_page.email.set('invalid')
+    image_share_page.message.set('hello')
+
+    image_share_page = image_share_page.share!.as_a(PageObjects::Images::SharePage)
+    assert_equal 'Email is invalid', image_share_page.email.error_message
+
```
rmacklin: At this point, you should also assert that the email input contains the previously entered value.

test/flows/share_image_flow_test.rb
```diff
@@ -0,0 +1,74 @@
...
+    image_share_page.email.set('valid@google.com')
+    images_index_page = image_share_page.share!
+    assert_equal 'Message sent.', images_index_page.flash_message(:success)
```
rmacklin: There is a lot of duplication in this test and the prior test. I think only one of them needs to extensively test the SharePage, while the other can just test the simple go-right case.

test/page_objects/images/image_card.rb
```diff
@@ -9,6 +9,21 @@ def tags
...
+
+      def delete
+        node.click_on('delete')
+        yield node.session.driver.browser.switch_to.alert
+      end
+
+      def delete_and_confirm!
+        delete(&:accept)
+        window.change_to(IndexPage)
+      end
```
rmacklin: Moving these methods out of `ShowPage` and into `ImageCard` is an independent refactor that can/should be in its own commit.

test/forms/share_mail_form_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  test 'email validation should reject invalid addresses' do
+    invalid_addresses = %w(user@example,com user_at_foo.org user.name@example.
+                           foo@bar_baz.com foo@bar+baz.com foo@bar..com)
+    invalid_addresses.each do |invalid_address|
+      form = ShareMailForm.new(email: invalid_address)
+      assert_predicate form, :invalid?
+      assert_equal 'is invalid', form.errors[:email].first
+    end
+  end
+end
```
rmacklin: Did you verify that the tests inside the new `forms` directory get run when you execute `rake test`?

test/controllers/images_controller_test.rb
```diff
@@ -112,4 +112,76 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+
+    assert_equal 'Check out this image', share_email.subject
+    assert_equal 'friend@example.com', share_email.to[0]
+
+    [:text_part, :html_part].each do |part|
+      email_body = share_email.send(part).body.to_s
+      assert email_body.include?(image.link), 'Email body does not contain image link'
+      assert email_body.include?("Your friend's message:"), 'Email body does not contain required message'
+      assert email_body.include?('this is a test'), 'Email body does not contain optional message'
+      assert email_body.include?('http://localhost:3000/'), 'Email body does not contain a link to app'
```
rmacklin: There is an `assert_includes` method that you should take advantage of, which has a built in helpful failure message, allowing you to get rid of these custom failure messages (here and in the rest of the PR)

rmacklin:
It's getting close. I left a few more comments. Additionally, regarding the commit order, I would move the app.json changes for sendgrid and the action_mailer config earlier in the history so that they are grouped together, since all of that is just getting email delivery configured for your application, prior to actually using it in a feature.

test/controllers/images_controller_test.rb
lsq921224: well actually I had tried that, but it did not behave the way i expected as it always gave me true no matter what i put for the message. I looked it up and the document says

> assert_includes(collection, obj, msg = nil)
> Fails unless collection includes obj.

so I thought it checks if an object is included in one collection, it might not work for checking if a string contains another string. but since you mention it, I will double check to see if it works.

rmacklin: It'll work for anything that implements `include?`, which [`String` does](http://ruby-doc.org/core-2.2.0/String.html#method-i-include-3F). And what I was saying is you don't need to pass a custom message if you use `assert_includes` because it will give a nice failure message automatically.

test/forms/share_mail_form_test.rb
lsq921224: Yes. If I add `raise` in one of these tests, `rake test` will throw runtime error

rmacklin: Ok cool 👍

lsq921224:
thanks for the comments. I have updated the PR.

test/test_helper.rb
```diff
@@ -2,6 +2,7 @@
 require File.expand_path('../../config/environment', __FILE__)
 require 'rails/test_help'
 require 'mocha/mini_test'
+require_relative '../app/validators/email_format_validator'
```
rmacklin: Is this only needed for the `email_format_validator_test`? If so, let's move the require there.

test/models/user_test.rb
```diff
@@ -36,24 +36,6 @@ def setup
...
-
-  test 'email validation should reject invalid addresses' do
-    invalid_addresses = %w(user@example,com user_at_foo.org user.name@example.
-                           foo@bar_baz.com foo@bar+baz.com foo@bar..com)
-    invalid_addresses.each do |invalid_address|
-      @user.email = invalid_address
-      assert_predicate @user, :invalid?
-    end
-  end
-
```
rmacklin: Removing these makes sense because you've moved them to the validator test. However, now we don't have any test that ensures `User` does not accept invalid emails. We can add a simple test here that just checks the `email` attribute with an invalid and a valid email address and asserts that the model is `invalid?` and `valid?`, respectively. This tests that `User` is using the `EmailFormatValidator`, without repeating the extensive testing of the regex (which belongs in the `EmailFormatValidatorTest` where you moved it).

rmacklin:
Great job! I just left two additional inline comments. I've got one more thing to add:

You've done a really good job creating atomic commits. There is a natural extension of atomic commits: atomic branches. This branch currently actually has 3 separate deliverables:

1. Extract the email format validation to a custom validator and refactor `User` to use it
2. Configure the application to be able to send emails
3. Add the ability to share images by email

It's a best practice to deliver these separate deliverables, you guessed it, separately. Why is that so? Well, for many reasons, but here are a few examples:

- By delivering (1) separately, those changes won't have to wait on this feature being merged, which a) reduces the possibility of merge conflicts in the `User` model if other changes need to be made in that model and b) allows the `EmailFormatValidator` to be reused for other models prior to the "share image" branch being merged.
- By delivering (2) separately, you can fully test the email setup in your staging and production environments without it being tied to the "share image" feature; moreover, other features that require email deliver can then be built without waiting for the "share image" branch to be merged.
- By doing these three things in 3 PRs, each PR is smaller in scope and thus, more easily testable, more easily reviewable, and less likely to introduce bugs. Furthermore, if the sharing feature had a bug, you could revert the "share image" PR without reverting the ability to send emails or the custom email format validator

With that said, can you extract (1) and (2) from this PR into their own PRs?

rmacklin:
Regarding
> 2. Configure the application to be able to send emails

and
> you can fully test the email setup in your staging and production environments without it being tied to the "share image" feature

You may be wondering how this can be tested without a feature that actually sends emails. The answer is: your good friend, the rails console. You can run the console on your review app to create a test mailer and send yourself an email from the review app. Once that's confirmed to work, you can merge the PR and do the same on your staging application, and after verifying it there you can promote staging to production and finally test it there.

rmacklin:
Lastly, if you're wondering why we didn't create separate stories for each of these deliverables, the reason is that this mimics real development. It's a natural part of the process to create a user story, start implementing it, and then realize that you can split out a few separate deliverables and merge some changes sooner than others. If we had made these separate stories from the beginning, you wouldn't get to develop this skill yourself.

lsq921224:
> Lastly, if you're wondering why we didn't create separate stories for each of these deliverables, the reason is that this mimics real development. It's a natural part of the process to create a user story, start implementing it, and then realize that you can split out a few separate deliverables and merge some changes sooner than others. If we had made these separate stories from the beginning, you wouldn't get to develop this skill yourself.

haha, I was about to ask this question but I knew you would have answered this XD

lsq921224:
I have divided this branch into smaller ones and delivered them. I have also verified that this review app works. Is it good to merge it now?

rmacklin:
Great work! I noticed that you have a typo in your last commit message. We can merge this after that's fixed.

rmacklin:
Great work! I noticed that you have a typo in your last commit message. Also, I noticed that the second commit references a view template that hasn't yet been created. We can move the creation of the `images/share/form` template to its own commit and move it prior to that so that the tests will pass at each commit.

lsq921224:
wow. that's a really good catch.. I have moved the share form template prior to 'Add share action' and fixed the typo in the last commit.

rmacklin:
Looks good! Awesome job on this 👍

PR #44, opened 2016-08-29
Add validation on tag list

### LINKED ISSUE #12: Required Tag Association
#### As a user I want all images to have at least one tag.

__Story__:

The tag feature has been a hit. Our users love being able to filter images by
cats, SR-71s, and "AppFolio tshirts" to name a few. In fact, we have discovered
that images without any tags receive significantly less views than those with
tags. As a result, we have decided to require all images to have at least one
tag.

__Acceptance criteria__:
- [ ] I cannot save an image without adding at least one tag.
- [ ] An error message is associated with the appropriate input field when a
tag is not provided.
- [ ] All images have at least one tag.

__Discussion Topic__:
- [ ] What do you do about existing images that have no tags?

__Dependencies__:
- Image Tags

lsq921224:
Closes #12


test/controllers/images_controller_test.rb
```diff
@@ -56,7 +65,7 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
   end
 
   test 'should display newly added image on index page' do
-    image = Image.create!(link: 'http://www.w3schools.com/css/trolltunga.jpg', tag_list: '')
+    image = Image.create!(link: 'http://www.w3schools.com/css/trolltunga.jpg', tag_list: 'tag')
```
pololee: Remember when we pair, we refactor the image model test and extract this valid params into private method. Do you think we could have a similar refactor in this controller test?

lsq921224: yes!

test/controllers/images_controller_test.rb
```diff
@@ -10,33 +10,40 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
     assert_difference 'Image.count' do
-      post images_path, params: { image: { link: 'http://www.w3schools.com/css/trolltunga.jpg',
-                                           tag_list: '' } }
+      post images_path, params: { image: image_valid_params }
     end
     assert_equal 'Image created successfully.', flash[:success]
     assert_redirected_to image_path(Image.last)
   end
 
   test 'should be able to create a valid image with tags' do
```
pololee: It looks like this test is the same as the test `should be able to create a valid image` above. Should we just keep just one?

test/controllers/images_controller_test.rb
```diff
@@ -10,33 +10,40 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    assert_select '.form-control-feedback', 'Link is not a valid url'
+  end
+
+  test 'should not be able to create an image without tag_list' do
+    assert_no_difference 'Image.count' do
+      post images_path, params: { image: image_valid_params.merge(tag_list: '') }
+    end
+    assert_response :unprocessable_entity
+    assert_select 'form[action=?]', images_path
+    assert_select '.form-control-feedback', 'Tag list should have at least one tag'
```
pololee: It feels like these `assert_select` are not necessary. They have been covered in your flow test.

lsq921224: I think it’s pretty cheap to do an `assert_select`, so even though it’s covered in flow test, I think it’s ok to put it in controller test to make sure i have the expected content

PR #49, opened 2016-09-08
Modify tag list

### LINKED ISSUE #13: Modify Tag List
#### As a user I want to edit the tag list for an image.

__Story__:

Image tags are awesome -- our user's love them! Tags help us make sense of the
plethora of images added to our system. But how annoying is it when our users
make spelling mistakes in their image tags? Super annoying! How awesome, and
useful would it be to allow for tags to be edited after an image is added?
Super awesome, and useful!

__Acceptance criteria__:
- [ ] There is a link to edit the tags for an image.
- [ ] The image whose tags are to be modified is displayed on the edit page.
- [ ] On the edit page, I can change the tags to any valid tag list.
- [ ] After saving the edit page with valid tags, my changes are visible
  throughout the rest of the application.

__Discussion__:
- [ ] Should an image's URL be editable?

__Dependencies__:
- Image Tags

lsq921224:
Closes #13


app/views/images/_form.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= simple_form_for @image do |f| %>
```
rmacklin: Prefer using local variables rather than instance variables in partials; it makes them more easily reusable/less coupled to a particular view context.

app/views/images/_form.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= simple_form_for @image do |f| %>
-  <%= f.input :link %>
+  <%= f.input :link, disabled: (true if controller.action_name == 'edit') %>
```
rmacklin: Along the same lines as my [previous comment](https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/49#discussion_r77931705), you should avoid having your partial inspect the environment in which its rendered. Instead, invert the dependency such that the callers who render this partial explicitly supply the information that the partial requires.

app/assets/stylesheets/_bootstrap-override.scss
```diff
@@ -1,3 +1,7 @@
+.btn {
+  margin-right: 4px;
+}
```
rmacklin: This may be fine, but I just want to ask if this is something you truly want to apply to _all `.btn` elements throughout the application_ or if it was only added for some subset of these elements. If the latter, targeting this style more specifically would be preferred. As usual for CSS, I'll defer to @aaronpanch's opinion on the matter.

test/controllers/images_controller_test.rb
```diff
@@ -199,4 +199,34 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  test 'should not be able to update a valid image with invalid tags' do
+    image = Image.create!(link: 'http://a.com/a.jpg', tag_list: 'tag')
+
+    put image_path(image), params: { image: { tag_list: '' } }
+    assert_response :unprocessable_entity
+
+    assert_equal %w(tag), image.reload.tag_list, 'Tag list is modified'
+  end
+
+  test "image's url should not be modified even if a user does an explicit http request" do
```
rmacklin: Great job testing this case!

test/controllers/images_controller_test.rb
```diff
@@ -199,4 +199,34 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    image = Image.create!(image_valid_params)
+
+    put image_path(image), params: { image: { link: 'http://google.com/a.jpg', tag_list: 'new, tag' } }
```
rmacklin: The way this test is currently written is a bit brittle. If someone were to update `image_valid_params` to have the `link` set to `'http://google.com/a.jpg'`, this test would no longer ensure that the URL wasn't updated. If we are explicit about the URL the image is created with we can avoid this.

rmacklin: Following up on this, I found the [`image_valid_params` method](https://github.com/AppFolioOnboarding/image-sharer-lsq921224/blob/7aa9c390ff8e5848650b2c4b709601684afcb02d/test/test_helper.rb#L39-L44) which I had not seen before. I see that method was added to DRY some of the `Image.create!` calls in your tests such that adding a new required attribute wouldn't require updating each of those. That's good but I think it would be better to extract the image creation itself into a private method that accepts arguments which would override the defaults:
```ruby
def create_image(link: 'http://google.com/a.jpeg', tag_list: 'tag, tag1, tag2')
  Image.create!(link: link, tag_list: tag_list)
end
```
This makes it easy to pass an explicit `link` and still have the valid default tags added for you.

Additionally, defining those methods directly in `ActiveSupport::TestCase` is an antipattern because it will make those methods available in all your tests, even where they aren't needed (e.g. none of your unit tests need to know how to log a user in). To avoid that, you can create modules to contain different test helper methods and `include` those modules in _only_ the tests that need them.

test/controllers/images_controller_test.rb
```diff
@@ -199,4 +199,34 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+
+    assert_equal 'Image updated successfully.', flash[:success]
+    assert_equal %w(tag new), image.reload.tag_list, 'Tag list is not updated'
+  end
+
+  test 'should not be able to update a valid image with invalid tags' do
+    image = Image.create!(link: 'http://a.com/a.jpg', tag_list: 'tag')
+
+    put image_path(image), params: { image: { tag_list: '' } }
+    assert_response :unprocessable_entity
```
rmacklin: What else can be asserted about the response?

app/controllers/images_controller.rb
```diff
@@ -1,5 +1,5 @@
 class ImagesController < ApplicationController
-  before_action :find_image, only: [:destroy, :share, :send_mail]
+  before_action :find_image, only: [:destroy, :share, :send_mail, :update]
```
weixing2014: It seems functions `edit` and `show` need to call `find_image` as the first step of each function as well.

app/views/images/_form.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= simple_form_for @image do |f| %>
+  <%= f.input :link, disabled: (true if controller.action_name == 'edit') %>
```
weixing2014: If you really want to disable the link only if the action name is `edit`, maybe below is slightly easier to read:
```
<%= f.input :link, disabled: controller.action_name.eql? 'edit' %>
```

test/page_objects/images/image_card.rb
```diff
@@ -14,6 +14,11 @@ def go_to_share_page!
         window.change_to(SharePage)
       end
 
+      def go_to_edit_page!
```
weixing2014: I think we usually name this function `edit!` in APM.
@rmacklin should know more about it.

app/controllers/images_controller.rb
lsq921224: Well, there's a slight difference. `edit` and `show` calls `find`: 
`    @image = Image.find(params[:id])
`
but the others call `find_by`:
`    @image = Image.find_by(id: params[:id])
`
I use `find` for `edit` and `show` because rails will then take care of the `RecordNotFound` case for me.
the reason why I did not DRY `edit` and `show` is because if I do so, these two functions will have nothing in their body, and I think it reduces the readability. But since you mention it, I guess I will do it as well.

weixing2014: @lsq921224 Thanks for explaining between `find` and `find_by`! Also I did some research and it seemed there are some different opinions about [using `before_action`](http://craftingruby.com/posts/2015/05/31/dont-use-before-action-to-load-data.html). Then I think it's fine to leave it as is. Thanks!

lsq921224:
@rmacklin @weixing2014 I think I have addressed your comments except https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/49#discussion_r77933018, for which I will open a separate PR to refactor my tests.

app/views/images/_form.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= simple_form_for image do |f| %>
-  <%= f.input :link %>
+  <%= f.input :link, disabled: controller.action_name.eql?('edit') %>
```
rmacklin: This comment still applies: https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/49#discussion_r77931893

> you should avoid having your partial inspect the environment in which it's being rendered. Instead, invert the dependency such that the callers who render this partial explicitly supply the information that the partial requires.

app/views/images/_form.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= simple_form_for image do |f| %>
+  <%= f.input :link, disabled: controller.action_name.eql?('edit') %>
```
rmacklin: This comment still applies: https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/49#discussion_r77931893

> you should avoid having your partial inspect the environment in which it's being rendered. Instead, invert the dependency such that the callers who render this partial explicitly supply the information that the partial requires.

lsq921224: Sorry that I missed it because Xing had another comment on that line. I have updated it.

app/views/images/_form.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= simple_form_for image do |f| %>
+  <%= f.input :link, disabled: disable_link %>
```
rmacklin: Minor but `disable_link_input` would be a little more clear (since `disable_link` could be misinterpreted as disabling a link on the page).

rmacklin: That said, good job making this change in response to https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/49#discussion_r78112197. However, I wanted to discuss an alternative. Do you think it's necessary to show this disabled input on the edit page, given that it's not editable, and we know which image we're editing because a preview is displayed on the edit page?

rmacklin: If you think we can omit the `link` input on the edit page, I can use this opportunity to show you a cool feature of rendering partials that makes them more flexible: partial layouts.

Of course, one way to omit the `link` input would be with something like:
```erb
<% if show_link_input %>
  <%= f.input :link, disabled: disable_link %>
<% end %>
```

However, this makes the partial very inflexible. All you can do is either render the `link` input or not. In our case, we actually want to be able to render the `link` input or render a preview of the image. We can achieve this flexibility using partial layouts.

If you replace the `link` input with a `yield` in your partial, like:
```erb
<%= simple_form_for image do |f| %>
  <% yield f if block_given? %>

  <%= f.input :tag_list, input_html: { value: image.tag_list.join(',') } %>
  <div class="form-group row">
    <%= f.button :submit, class: 'btn-primary' %>
  </div>
<% end %>
```
You can render it as a partial layout. For the new page you'd pass a block that renders the `link` input:
```erb
<%= render layout: 'images/form', locals: { image: @image } do |form| %>
  <%= form.input :link %>
<% end %>
```
And for the edit page, you can pass a block that renders the image preview:
```erb
<%= render layout: 'images/form', locals: { image: @image } %>
  <div class="form-group row">
    <%= link_to image_path(@image), class: 'offset-sm-2' do %>
      <%= image_tag(@image.link, class: 'share-form-image') %>
    <% end %>
  </div>
<% end %>
```

This approach lets you reuse the pieces of the image form that are actually shared, while giving you the flexibility to insert anything else you want from the template that's rendering this partial.

rmacklin: cc @weixing2014

weixing2014: Thanks, @rmacklin !

lsq921224: @rmacklin I really like this idea of  `<% yield f if block_given? %>`, but for some reason it's not showing the block I supply to it. i have changed my code to https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/49#discussion_r78235833 exactly. I am trying to figure out why, but it would be great if you can give me some clue on that.

rmacklin: It looks like a typo in my comment,
`<% yield f if block_given? %>`
should be
`<%= yield f if block_given? %>`
so that the block is actually rendered. Does that fix it?

lsq921224: yes, it works!

app/views/images/new.html.erb
```diff
@@ -1,7 +1,3 @@
-<%= simple_form_for @image do |f| %>
-  <%= f.input :link %>
-  <%= f.input :tag_list, input_html: { value: @image.tag_list.join(',') } %>
-  <div class="form-group row">
-    <%= f.button :submit, class: 'btn-primary' %>
-  </div>
+<%= render layout: 'images/form', locals: { image: @image } do |form| %>
+  <%= form.input :link %>
```
rmacklin: Nice!

app/views/images/edit.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= render layout: 'images/form', locals: { image: @image } do %>
+<div class="form-group row">
+  <%= link_to image_path(@image), class: 'offset-sm-2' do %>
+    <%= image_tag(@image.link, class: 'share-form-image') %>
+  <% end %>
+</div>
```
rmacklin: Minor: this should be indented a level

rmacklin: Minor: this should be indented a level. Other than that, nice!

rmacklin:
> I think I have addressed your comments except https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/49#discussion_r77933018, for which I will open a separate PR to refactor my tests.

I saw that PR (#51) and left comments there. Whichever is merged first will require changes in the other.

lsq921224:
I fixed this https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/49#discussion_r78458915 and I'm merging this PR, then I will make corresponding change in https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/51

PR #50, opened 2016-09-09
Share image part2

### LINKED ISSUE #10: Share Image

lsq921224:
Part 2 of image sharing

Closes #10


app/views/layouts/application.html.erb
```diff
@@ -4,7 +4,6 @@
     <title>Image Sharer</title>
     <%= csrf_meta_tags %>
     <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
```
rmacklin: You should also remove the turbolinks-track data attribute from this line

rmacklin: Also, this commit can be split further into atomic commits.

app/assets/javascripts/images/index/helper.js
```diff
@@ -0,0 +1,5 @@
+export default function setFlashMessage(type, message) {
```
rmacklin: The name of this file is `helper.js` which implies it is some kind of generic helper. But it's just a `setFlashMessage` function. We should rename the file to match its default (and only) export: `set_flash_message.js`

app/assets/javascripts/images/show/index.js
```diff
@@ -0,0 +1 @@
+require('../index/main')
```
rmacklin: Rather than reaching into another page's files, can we move/rename `index/main` so it's clear that it's reusable?

app/assets/javascripts/images/index/share_modal.js
```diff
@@ -0,0 +1,41 @@
...
+import setFlashMessage from './helper';
+
+class ShareModal {
+  constructor(modalSelector) {
+    this.$modal = $(modalSelector);
+  }
+
+  setupModal({imageId, imageUrl}) {
+    this.$modal.find('#js_image_link').attr('href', `/images/${imageId}`);
+    this.$modal.find('#new_share_mail_form').attr('action', `/images/${imageId}/share`);
```
rmacklin: It's not ideal to hardcode these paths in our javascript. We can avoid this if we store the "template" for an "image share path" and use that template to construct the real image share path here.

(Hint, you'd have something like `sharePathTemplate.replace('ID_PLACEHOLDER', imageId)`)

app/assets/javascripts/images/index/share_modal.js
```diff
@@ -0,0 +1,41 @@
...
+      const modalTrigger = event.relatedTarget;
+      this.setupModal(modalTrigger.dataset);
+    });
+
+    this.$modal.on('ajax:error', (event, xhr) => {
+      if (xhr.status === 404) {
+        this.$modal.modal('hide');
+        window.location.replace("/");
+      } else if (xhr.status === 422) {
+        this.$modal.find('.modal-body').html(xhr.responseText);
```
rmacklin: We should use a JSON protocol with a specific key for this rather than just relying on it being the response text.

app/assets/javascripts/images/index/share_modal.js
```diff
@@ -0,0 +1,41 @@
...
+      } else {
+        console.log(xhr.status);
```
rmacklin: Do we want to ship this line to production?

app/assets/javascripts/images/index/share_modal.js
```diff
@@ -0,0 +1,41 @@
...
+        this.$modal.modal('hide');
+        setFlashMessage('error', 'Problems sending message.')
```
rmacklin: Since this is an unexpected, possibly transient, server error, consider a different user experience. Instead of hiding the modal and displaying this flash message, we could show a browser alert with the additional text "Please try again"

app/controllers/images_controller.rb
```diff
@@ -71,4 +70,44 @@ def mail_form_params
   def find_image
     @image = Image.find_by(id: params[:id])
   end
+
+  def deliver_mail(image:, form:)
```
rmacklin: This method is called `deliver_mail` but it does more than just delivering mail.  Does `respond_to_message_sent` need to be part of this method?

app/controllers/images_controller.rb
```diff
@@ -71,4 +70,44 @@ def mail_form_params
...
+    ).deliver_now!
+
+    respond_to_message_sent
+  end
+
+  def respond_to_message_sent
+    respond_to do |format|
+      flash[:success] = 'Message sent.'
+      format.html { redirect_to images_path }
+      format.js { head :ok }
```
rmacklin: As mentioned in my earlier comment, lets expect the client to request `json` rather than `js`

test/page_objects/images/share_form.rb
```diff
@@ -0,0 +1,25 @@
...
+          element :email
+          element :message
+        end
+      end
+
+      def url
+        node.find('img')[:src]
+      end
+
+      def share(email: nil, message: nil, remote: true)
```
rmacklin: It'd be better to provide two different methods than to vary the behavior based on this `remote` parameter.

rmacklin:
Are there separate deliverables in this branch that could be broken off into their own PRs?

rmacklin:
I'll review further once the initial comments are addressed. Great job so far 👍

app/assets/javascripts/images/index/share_modal.js
lsq921224: ok.. then I can actually get rid of the set_flash_message.js..

app/views/layouts/application.html.erb
lsq921224: I split this into two commits: "remove turbolinks" and "add named yield for page-specific js and move js into end of html body". Should I split the latter one further?

app/assets/javascripts/images/index/share_modal.js
rmacklin: I think you should use `setFlashMessage` for the success case - I don't think we need to do a full page load in that case.

app/views/layouts/application.html.erb
rmacklin: Personally, I would, because those two operations serve different purposes, but I won't force you to split it if you don't want to

lsq921224: Ok, i will do so then. I just want to know the best practice..

lsq921224:
I have updated this PR. @rmacklin please take a look.

app/views/images/share/_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= simple_form_for form, url: share_image_path(image || { id: 'ID_PLACEHOLDER' }), remote: remote do |f| %>
+  <div class="form-group row">
+    <%= link_to image_path(image || { id: 'ID_PLACEHOLDER' }), class: 'offset-sm-2', id: 'js_image_link' do %>
+      Image to share
+      <%= image_tag(image.present? ? image.link : '', class: 'share-form-image', id: 'js_image_tag') %>
```
rmacklin: Could use `image.try(:link).to_s` here

app/views/images/share/_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= simple_form_for form, url: share_image_path(image || { id: 'ID_PLACEHOLDER' }), remote: remote do |f| %>
+  <div class="form-group row">
+    <%= link_to image_path(image || { id: 'ID_PLACEHOLDER' }), class: 'offset-sm-2', id: 'js_image_link' do %>
```
rmacklin: `id`s do not need a `js` prefix. But we may want to use a more specific `id` like `share_form_image_link`

app/views/images/share/_form.html.erb
rmacklin: Similar to https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/50#r78895812, this consider something like `share_form_image_preview for this `id`

app/assets/javascripts/images/shared/set_flash_message.js
```diff
@@ -0,0 +1,5 @@
+export default function setFlashMessage(type, message) {
+  $('#flash_messages').html(
+    `<div class=\"alert alert-${type}\">${message}</div>`
+  );
+}
```
rmacklin: This is applicable to the whole application, not just the `images` views, so I don't think it belongs under `images/shared`

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,44 @@
...
+  attachEventHandlers() {
+    this.$modal.on('show.bs.modal', event => {
+      const modalTrigger = event.relatedTarget;
+      this.setupModal(modalTrigger.dataset);
+    });
+
+    this.$modal.on('ajax:error', (event, xhr) => {
+      if (xhr.status === 404) {
+        this.$modal.modal('hide');
+        window.location.replace("/");
```
rmacklin: Minor, but it seems the rest of this file is using single quoted strings.

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,44 @@
...
+      } else if (xhr.status === 422) {
+        this.$modal.find('.modal-body').html(xhr.responseJSON.share_form);
+      } else {
+        this.$modal.modal('hide');
```
rmacklin: Since we are telling them to try again, I don't think we should hide the modal

app/assets/javascripts/images/share_modal.js
```diff
@@ -0,0 +1,44 @@
...
+        alert('Please try again');
```
rmacklin: Consider "Sorry, something went wrong. Please try again." or "Sorry, an unexpected error occurred. Please try again."

app/controllers/images_controller.rb
```diff
@@ -87,4 +87,50 @@ def mail_form_params
...
+      format.json { head :not_found }
+      format.html { redirect_to images_path }
+    end
+  end
+
+  def respond_to_invalid_form(image:, form:)
+    share_form_remote = render_to_string(
+      partial: 'images/share/form',
+      locals: { form: form, image: image, remote: true }
+    )
```
rmacklin: What's the reason for putting this in this method if it's only used in the `json` case? Seems like it could be moved to the `render_share_form_error_json` case.

app/controllers/images_controller.rb
```diff
@@ -87,4 +87,50 @@ def mail_form_params
...
+    success_msg = 'Message sent.'
+    respond_to do |format|
+      format.json { render json: { msg: success_msg } }
+      format.html { redirect_to images_path, success: success_msg }
+    end
+  end
+
+  def respond_to_no_image
+    respond_to do |format|
+      flash[:error] = 'Cannot share this image. It may have been deleted.'
```
rmacklin: Does this need to be inside the `respond_to` block?

app/views/images/_image.html.erb
```diff
@@ -10,7 +10,13 @@
         <% end %>
       <% end %>
       <div class="image-card__buttons">
-        <%= link_to 'share', share_image_path(image), class: 'btn btn-primary' %>
+        <%= link_to 'share', share_image_path(image), class: 'btn btn-primary',
+                    data: {
+                      toggle: 'modal',
+                      target: '#share_modal',
+                      'image-id': image.id,
+                      'image-url': image.link
```
rmacklin: I think if you use
```erb
image_id: image.id,
image_url: image.link
```
it will automatically get converted and work the same.

test/page_objects/images/share_form.rb
```diff
@@ -0,0 +1,29 @@
...
+      def share_remote(email: nil, message: nil)
+        self.email.set email if email.present?
+        self.message.set message if message.present?
+        node.click_button('Share')
+      end
+
+      def share!(email: nil, message: nil)
+        share_remote(email: email, message: message)
+        window.change_to(IndexPage, self.class)
+      end
```
rmacklin: Seems this method only applies to the `SharePage`, so it can be moved there.

test/page_objects/images/share_form.rb
```diff
@@ -0,0 +1,29 @@
...
+          element :email
+          element :message
+        end
+      end
+
+      def url
+        node.find('img')[:src]
+      end
+
+      def share_remote(email: nil, message: nil)
```
rmacklin: Consider just calling this `share` - `share_remote` doesn't have meaning to our users.

test/flows/share_image_flow_test.rb
```diff
@@ -4,18 +4,11 @@
 class ShareImageFlowTest < FlowTestCase
   include ImageCreation
 
-  test 'share an image from show page' do
+  test 'share an image from share page (no js case)' do
```
rmacklin: `(no js case)` isn't technically correct. You can still get to the page with JS enabled if you right click on the share button and select "Open in new tab"

test/flows/share_image_flow_test.rb
```diff
@@ -33,7 +26,7 @@ class ShareImageFlowTest < FlowTestCase
     assert_equal 'Message sent.', images_index_page.flash_message(:success)
   end
 
-  test 'share an image from index page' do
+  test 'share an image from index page using show modal' do
```
rmacklin: "using modal from show page" would be more clear

app/controllers/images_controller.rb
lsq921224: that's what I did. I moved this out because if this is called inside `format.json` block, I will get missing template error as rails will be looking for  `images/share/form.json.erb` partial

test/flows/share_image_flow_test.rb
```diff
@@ -58,13 +86,33 @@ class ShareImageFlowTest < FlowTestCase
...
+    create_image(link: puppy_url, tag_list: 'superman, cute')
+
+    images_index_page = PageObjects::Images::IndexPage.visit
+    image_to_share = images_index_page.images.find do |image|
+      image.url == puppy_url
+    end
+
+    ImagesController.any_instance.stubs(:send_mail).raises(RuntimeError, 'an error')
+
+    suppress(RuntimeError) do
```
rmacklin: Are you sure this is necessary? The `RuntimeError` should be happening in a different process than this test.

test/flows/share_image_flow_test.rb
```diff
@@ -58,13 +86,33 @@ class ShareImageFlowTest < FlowTestCase
...
+      # dangerous code here
+      image_to_share.share do |image_share_modal|
+        image_share_modal.email.set('valid@google.com')
+        image_share_modal.share_remote
+        accept_alert
```
rmacklin: It would be better to make an assertion about the alert message.

test/flows/share_image_flow_test.rb
```diff
@@ -58,13 +86,33 @@ class ShareImageFlowTest < FlowTestCase
...
+    image_to_share.share do |image_share_modal|
+      image_share_modal.email.set('valid@google.com')
+      image_share_modal.share_remote
+    end
 
-    image_share_page.email.set('valid@google.com')
-    image_share_page.message.set('hello')
-
-    images_index_page = image_share_page.share!
     assert_equal 'Cannot share this image. It may have been deleted.', images_index_page.flash_message(:error)
```
rmacklin: This `images_index_page` is a stale reference because the client reloaded the page, which means this line is flaky. You'll want to use
```ruby
images_index_page =
  AePageObjects.browser.current_window.change_to PageObjects::Images::IndexPage
```
to get a reference to the new (not stale) page

test/flows/share_image_flow_test.rb
rmacklin: Also, you could `unstub` the method and submit the form again to assert that the error can be recovered from.

rmacklin:
I'm checking your review app. The message field looks very small:

screen shot 2016-09-14 at 11 52 36 pm

rmacklin:
Also, I found a bug:
peterbug

This is a great opportunity to test-drive the bug fix. That is:

1. Write a test for this case which will fail with your current code
2. Fix the bug
3. Observe the the test now passes

test/flows/share_image_flow_test.rb
```diff
@@ -42,13 +35,48 @@ class ShareImageFlowTest < FlowTestCase
...
+      image_share_modal.email.set('valid@google.com')
+      image_share_modal.message.set('hello')
+
+      image_share_modal.share_remote
+    end
+
+    assert_equal 'Message sent.', image_show_page.flash_message(:success)
+  end
+
   test 'share an image that does not exist' do
```
rmacklin: There should still be a test for this case that doesn't use the modal

test/flows/share_image_flow_test.rb
```diff
@@ -58,13 +86,33 @@ class ShareImageFlowTest < FlowTestCase
       image.url == puppy_url
     end
 
-    image_share_page = image_to_share.go_to_share_page!
     Image.last.destroy!
+    image_to_share.share do |image_share_modal|
+      image_share_modal.email.set('valid@google.com')
+      image_share_modal.share_remote
+    end
```
rmacklin: This is flaky because `share` [yields to the block](https://github.com/AppFolioOnboarding/image-sharer-lsq921224/blob/5cc23507268a04e7a674d02715e1b22a778323c5/test/page_objects/images/image_card.rb#L16) and then [calls `share_modal.wait_until_hidden`](https://github.com/AppFolioOnboarding/image-sharer-lsq921224/blob/5cc23507268a04e7a674d02715e1b22a778323c5/test/page_objects/images/image_card.rb#L17). But by the time the block finishes, we have submitted the form which will receive a 404 response that triggers the client to [reload the index page](https://github.com/AppFolioOnboarding/image-sharer-lsq921224/blob/5cc23507268a04e7a674d02715e1b22a778323c5/app/assets/javascripts/images/share_modal.js#L28). So because the page reloads, `share_modal.wait_until_hidden` is actually referencing a stale element (`share_modal`). Because of this, we need to expose a second method in our page object interface which simply opens the share modal and returns it to the caller (rather than yielding it to a block):

```ruby
def open_share_modal
  node.click_on('share')
  share_modal = document.element(locator: '#share_modal', is: ShareModal)
  share_modal.wait_until_visible
  share_modal
end

def share(&block)
  share_modal = open_share_modal
  block.call(share_modal)
  share_modal.wait_until_hidden
end
```

The `share` method should be used the majority of the time. The `open_share_modal` method is only necessary for testing the case where you attempt to share a deleted image to avoid the flakiness of accessing a stale element that I described.

app/views/images/share/_form.html.erb
lsq921224: it took some time for me to understand how this works, but this is cool.

test/flows/share_image_flow_test.rb
lsq921224: thank you for the explanation.

test/flows/share_image_flow_test.rb
lsq921224: I don't know if there is a better way. Right now if I don't do this, this `RuntimeError` will actually fail me this test

> Minitest::UnexpectedError: RuntimeError: an error
>     
> Finished in 5.26050s
> 1 tests, 0 assertions, 0 failures, 1 errors, 0 skips

lsq921224:
I have addressed the comments except https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/50#r78898330 and https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/50#r78897273

rmacklin:
Can you address the comments in new commits and squash them only after I've seen them? If you squash them before I see them, it's harder for me to see the changes that were made to address the comments.

rmacklin:
Also, the bug fix does not handle everything. Look at this:

peterbug_2

app/controllers/images_controller.rb
rmacklin: Did you try explicitly specifying the template format? That should resolve the issue you mentioned.

lsq921224:
I have updated the PR with new commits.

rmacklin:
Here are two more bugs:

Clicking on the image in the share modal does not always take you to the correct image:

peterbug_3

Submitting the form with validations errors does not always rerender the correct image:

peterbug_4

There's also a third bug which I didn't record because it requires waiting for the email to be delivered (which takes a while since we're using the free plan). The bug is that if you share an image, and then immediately share a second image, the second email is still for sharing the first image.

Some takeaways from this:

1. During development you should do manual QA
2. When fixing a bug, look for the root cause, and try to fix that, rather than fixing a symptom.

rmacklin:
Just saw your message:

> I have updated the PR with new commits.

From looking at the new changes, I think those bugs should be fixed now, but the takeaways are still valid.

test/flows/share_image_flow_test.rb
```diff
@@ -141,17 +141,26 @@ class ShareImageFlowTest < FlowTestCase
...
     end
 
-    image_share_modal = image_to_share.open_share_modal
-    image_share_modal.email.set('thiswillbeblank@google.com')
-    image_share_modal.message.set('this will be blank')
-    image_share_modal.close
+    image_to_share.share do |image_share_modal|
+      image_share_modal.email.set('thiswillbeblank@invalid')
+      image_share_modal.message.set('this will be blank')
+      image_share_modal.share
```
rmacklin: Why are you calling
```rb
image_share_modal.share
image_share_modal.close
```
rather than just 
```rb
image_share_modal.close
```
?

test/flows/share_image_flow_test.rb
```diff
@@ -141,17 +141,26 @@ class ShareImageFlowTest < FlowTestCase
...
+      image_share_modal.close
+    end
+
+    images_index_page = AePageObjects.browser.current_window.change_to PageObjects::Images::IndexPage
```
rmacklin: Why are you doing this?

app/assets/javascripts/images/share_modal.js
```diff
@@ -23,7 +24,7 @@ class ShareModal {
     });
 
     this.$modal.on('hidden.bs.modal', () => {
-      this.$modal.find('.form-control').val('');
+      this.$modal.html(this.newModalHTML);
```
rmacklin: 👍  Nice job. Notice how this is a much more robust solution and gets at the "root cause" like I mentioned in https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/50#issuecomment-248192566

app/controllers/images_controller.rb
```diff
@@ -132,7 +127,12 @@ def render_share_form_error_html(image:, form:)
            status: :unprocessable_entity
   end
 
-  def render_share_form_error_json(share_form)
-    render json: { share_form: share_form }, status: :unprocessable_entity
+  def render_share_form_error_json(image:, form:)
+    share_form_remote = render_to_string(
+      'images/share/_form.html.erb',
+      locals: { form: form, image: image, remote: true },
+      layout: false
```
rmacklin: This works, but it's a bit hacky having to specify `layout: false` and a bit ugly having to use the full extension (`.html.erb`) inside the partial string. There's a way to do this that avoids both of those issues, which is to use the `formats` key:
```rb
share_form_html = render_to_string(
  'images/share/_form',
  formats: :html,
  locals: { form: form, image: image, remote: true }
)
```
(I also renamed the variable to `share_form_html` because that describes what it's storing. You could also rename the JSON key to that.)

test/flows/share_image_flow_test.rb
lsq921224: I added this because I want to test the error message will be erased as well, but in that commit I did not add corresponding assertions. I added them in the new fixup

lsq921224:
I have updated the PR in new commits.

test/flows/share_image_flow_test.rb
```diff
@@ -67,4 +91,80 @@ class ShareImageFlowTest < FlowTestCase
...
+      image_share_modal.share
+      assert_equal 'Email is invalid', image_share_modal.email.error_message
+      image_share_modal.close
+    end
+
+    image_to_share = images_index_page.images.find do |image|
+      image.url == cute_puppy_url
+    end
+
+    image_share_modal = image_to_share.open_share_modal
```
rmacklin: As mentioned in https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/50#discussion_r78908412, we should avoid using `open_share_modal` whenever we can (which is almost all cases) since it makes it possibly to write flaky tests. In this case, the usage comes just before the end of the test so you may think it's okay right now. And that's true: it's okay _right now_ -- but we want to be defensive against the future where someone could add lines after the end  of this test and because we've used `open_share_modal` there will be a possibility of accessing a stale `image_share_modal` element. Since it's super easy to avoid that, we should do it:
```rb
image_to_share.open_share_modal do |image_share_modal|
  assert_empty image_share_modal.email.value
  assert_empty image_share_modal.email.error_message
  assert_empty image_share_modal.message.value
  image_share_modal.close
end
```

test/flows/share_image_flow_test.rb
```diff
@@ -67,4 +91,80 @@ class ShareImageFlowTest < FlowTestCase
...
+
+    images_index_page = PageObjects::Images::IndexPage.visit
+    image_to_share = images_index_page.images.find do |image|
+      image.url == puppy_url
+    end
+
+    ImagesController.any_instance.stubs(:send_mail).raises(RuntimeError, 'an error')
+
+    assert_raise(RuntimeError) do
+      # dangerous code here
```
rmacklin: I don't think this comment is necessary.

test/flows/share_image_flow_test.rb
rmacklin: I'm still surprised by this - I don't think it used to work this way. But I like the change you made to use `assert_raises` instead of `suppress`. If I have time later I'll try to see if we can get rid of that too, but for now we can leave it.

test/flows/share_image_flow_test.rb
```diff
@@ -67,4 +91,80 @@ class ShareImageFlowTest < FlowTestCase
...
+    end
+
+    image_to_share = images_index_page.images.find do |image|
+      image.url == cute_puppy_url
+    end
+
+    image_share_modal = image_to_share.open_share_modal
+    assert_empty image_share_modal.email.value
+    assert_empty image_share_modal.email.error_message
+    assert_empty image_share_modal.message.value
```
rmacklin: We could also test the second bug mentioned in https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/50#issuecomment-248192566. To do that, we'd submit this form for sharing the second image, and assert that the image preview still shows the correct image.

Additionally, we could test the third bug mentioned in that comment as a new test which simply shares two images from the index page and asserts that the email our applications delivered included the appropriate image after each share.

test/flows/share_image_flow_test.rb
```diff
@@ -67,4 +91,80 @@ class ShareImageFlowTest < FlowTestCase
...
+
+        image_share_modal.share
+      end
+    end
+
+    images_index_page = AePageObjects.browser.current_window.change_to PageObjects::Images::IndexPage
+    assert_equal 'Message sent.', images_index_page.flash_message(:success)
+  end
+
+  test 'email and message should be reset when user hides share modal' do
```
rmacklin: Since this resets more than just the email and message inputs, consider a more general name like `share modal is reset after it is hidden`

test/flows/share_image_flow_test.rb
```diff
@@ -67,4 +91,80 @@ class ShareImageFlowTest < FlowTestCase
...
+    puppy_url = 'http://www.pawderosa.com/images/puppies.jpg'
+    cute_puppy_url = 'http://ghk.h-cdn.co/assets/16/09/980x490/landscape-1457107485-gettyimages-512366437.jpg'
+    create_image(link: puppy_url, tag_list: 'superman, cute')
+    create_image(link: cute_puppy_url, tag_list: 'puppy, cute')
+
+    images_index_page = PageObjects::Images::IndexPage.visit
+    image_to_share = images_index_page.images.find do |image|
```
rmacklin: Since we are sharing two images, we can give the variables distinct names, like `first_image_to_share` here and `second_image_to_share` on line 165.

rmacklin:
The new commits look good. I only had a few remaining comments on the flow tests. After those are addressed, please squash the fixups and then I will take one more holistic pass to make sure everything still looks good and then we can merge this.

As mentioned, really awesome job on this PR! I like that you went the extra mile by keeping the share page, which wasn't a requirement but makes the application more accessible.

lsq921224:
Thanks, I pushed the fix in a new commit and then rebased and squashed the fixups

app/controllers/images_controller.rb
```diff
@@ -87,4 +87,52 @@ def mail_form_params
...
+  end
+
+  def render_share_form_error_json(image:, form:)
+    share_form_html = render_to_string(
+      partial: 'images/share/form',
+      formats: :html,
+      locals: { form: form, image: image, remote: true }
+    )
+    render json: { share_form_html: share_form_html }, status: :unprocessable_entity
+  end
```
rmacklin: These controller changes should spawn accompanying controller test changes, but this PR doesn't have any controller test changes.

lsq921224:
I have added the controller tests corresponding to the controller changes in this fixup: https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/50/commits/21263888d3a615f268ef0b9652d4d7507ea929b2

test/controllers/images_controller_test.rb
```diff
@@ -203,6 +203,75 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+  test 'should not send email if address is invalid (json response)' do
+    image = create_image
+
+    assert_no_difference 'ActionMailer::Base.deliveries.size' do
+      post share_image_path(image), xhr: true, params: {
+        share_mail_form:
+          { image: image,
+            email: 'friend@example',
+            message: 'this is a test' },
+        format: :json
```
rmacklin: For readability it's preferred to align the closing brace with the line that had the opening brace:
```rb
post share_image_path(image), xhr: true, params: {
  share_mail_form: {
    image: image,
    email: 'friend@example',
    message: 'this is a test'
  },
  format: :json
}
```
(This applies to other test cases as well)

test/controllers/images_controller_test.rb
```diff
@@ -203,6 +203,75 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+      post share_image_path(image), xhr: true, params: {
+        share_mail_form:
+          { image: image,
+            email: 'friend@example.com',
+            message: 'this is a test' },
+        format: :json
+      }
+    end
+
+    assert_response :ok
```
rmacklin: What else should be asserted about the response?

test/controllers/images_controller_test.rb
```diff
@@ -203,6 +203,75 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+
+    [:text_part, :html_part].each do |part|
+      email_body = share_email.send(part).body.to_s
+      assert_includes email_body, image.link
+      assert_includes email_body, "Your friend's message:"
+      assert_includes email_body, 'this is a test'
+      assert_includes email_body, 'http://localhost:3000/'
+    end
+  end
+
```
rmacklin: Good job with these tests! I see that you've divided the share tests into HTML and JSON test cases. Personally, I prefer seeing the HTML test for one case (e.g. share with invalid email) and then immediately see the same test for the JSON request, so that it's easy to compare the two. If you don't want to change this that's okay, but I figured I'd share it anyway.

test/controllers/images_controller_test.rb
```diff
@@ -203,6 +203,75 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+      }
+    end
+
+    assert_response :unprocessable_entity
+
+    share_form_html = JSON.parse(response.body)['share_form_html']
+    assert_includes share_form_html, share_image_path(image)
+    assert_includes share_form_html, image.link
+    assert_includes share_form_html, 'share_mail_form_email'
+    assert_includes share_form_html, 'form-control'
```
rmacklin: Did you mean to use `form-control-feedback` here?

test/controllers/images_controller_test.rb
```diff
@@ -203,6 +203,75 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
...
+    assert_includes share_form_html, 'is invalid'
```
rmacklin: Did you mean to use `Email is invalid` here?

test/controllers/images_controller_test.rb
```diff
@@ -162,7 +162,7 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_select '.form-control-feedback', 'Email is invalid'
```
rmacklin: We can also assert that the input values were preserved, and that the image url and form action are correct in this test case, like you did in the JSON test case. As mentioned in my other comment, putting the JSON and HTML test cases next to each other helps compare the two which can uncover gaps in test coverage like this.

lsq921224:
I have made the corresponding changes to the images controller test in this new fixup commit: https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/50/commits/20ba7f8a99deb2b6e58db550caa9e59318b9a761

lsq921224:
@rmacklin thank you for guiding me to rewrite the git history to make it 
>more explicit, history-preserving, granular/atomic

I have reset/rebased this branch as well.

rmacklin:
Great job 👍

PR #51, opened 2016-09-09
Move test helpers to modules and refactor related tests

lsq921224:
Address https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/49#discussion_r77933018


test/test_helper.rb
```diff
@@ -26,7 +28,9 @@ def log_in_as(user, password: 'password', remember_me: 'on')
                                                password:    password,
                                                remember_me: remember_me } }
     end
+  end
 
+  module ImagesTest
```
rmacklin: A few things:
1. These don't belong inside the `ActiveSupport` namespace.
2. The modules shouldn't be named `___Test` because that suggests they are tests - which they aren't (they're test helpers).
3. These can be defined in their own files (e.g. under `test/test_helpers`)

test/controllers/users_controller_test.rb
```diff
@@ -1,6 +1,8 @@
 require 'test_helper'
 
 class ImagesControllerTest < ActionDispatch::IntegrationTest
```
rmacklin: This test class is named incorrectly

test/test_helper.rb
```diff
@@ -7,7 +7,9 @@ module ActiveSupport
   class TestCase
     # Setup all fixtures in test/fixtures/*.yml for all tests in alphabetical order.
     fixtures :all
+  end
 
+  module AuthenticationTest
```
rmacklin: It seems the `log_in_as` method is only used in the `SessionsControllerTest` so it can be moved to a private method there.

test/test_helper.rb
```diff
@@ -42,5 +46,13 @@ def image_valid_params
         tag_list: 'tag, tag1, tag2'
       }
     end
+
+    def new_image_with_default_params(link: 'http://google.com/a.jpeg', tag_list: 'tag, tag1, tag2')
+      Image.new(link: link, tag_list: tag_list)
+    end
+
+    def create_image_with_default_params(link: 'http://google.com/a.jpeg', tag_list: 'tag, tag1, tag2')
```
rmacklin: I would just call this method `create_image` since the default params only apply if you do not override them.

test/test_helper.rb
```diff
@@ -42,5 +46,13 @@ def image_valid_params
         tag_list: 'tag, tag1, tag2'
       }
     end
+
+    def new_image_with_default_params(link: 'http://google.com/a.jpeg', tag_list: 'tag, tag1, tag2')
```
rmacklin: I would just call this method `new_image` since the default params only apply if you do not override them.

test/test_helper.rb
rmacklin: Per (2), I would rename this module to `ImageCreation`

test/test_helper.rb
rmacklin: I would split this module into `UserCreation` with `create_user` and `LoggedInCheck` with `logged_in?` since those methods are needed in different places.

test/test_helper.rb
rmacklin: `image_valid_params` is now only used in `ImagesControllerTest`, so it can be moved to a private method there.

rmacklin: Similarly `create_images_with_tags` is only used in `TagsControllerTest` so it should be moved to a private method there. In general, things should only be extracted when they are reused in multiple places. Otherwise, a) there is an unnecessary level on indirection and b) you end up sharing things that don't need to be shared when you group it with other things that _are_ reused.

test/test_helper.rb
rmacklin: Also, it's a bit minor but I think `create_a_user` should be renamed to just `create_user` (for consistency with `create_image`).

lsq921224:
Thanks, I updated the code.

rmacklin:
Looks good, but it appears there are still a lot of tests that call `Image.create!` directly. I think these can be replaced to call the `create_image` factory. You could also add a `create_images` factory for creating multiple images for the tests that need more than one image created.

lsq921224:
I have a question about `create_images`, 
I did it in this way:
```
  def create_images(links_and_tags)
    Image.create!([links_and_tags.map { |link_and_tags| {
      link: link_and_tags[:link],
      tag_list: link_and_tags[:tag_list] }
    }])
 end
```
then I figure it's the same as:

```
  def create_images(links_and_tags)
    Image.create!(links_and_tags)
  end
```

then I don't see a point of having this method: unlike `create_image`, I don't really have default settings for `create_images`

rmacklin:
You can implement `create_images` such that there _are_ default values:
```ruby
def create_images(image_attributes_list)
  image_attributes_list.map(&method(:create_image))
end
```
`images_attributes_list` would be an array of hashes that optionally override any of the default keys, otherwise the defaults would be used for any non-overridden keys.

lsq921224:
Oh wow, this is amazing.

lsq921224:
I have updated this PR after #49 is merged.

test/controllers/images_controller_test.rb
```diff
@@ -101,9 +118,9 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
   end
 
   test 'should filter by the first tag, if the query contains 2 tags' do
-    image1 = Image.create!(link: 'http://www.w3schools.com/css/trolltunga.jpg', tag_list: 'tag,1')
-    image2 = Image.create!(link: 'http://www.w3schools.com/css/a.jpg', tag_list: 'tag,2')
-    image3 = Image.create!(link: 'http://www.google.com/a.jpg', tag_list: '3')
+    image1 = create_image(link: 'http://www.w3schools.com/css/trolltunga.jpg', tag_list: 'tag,1')
+    image2 = create_image(link: 'http://www.w3schools.com/css/a.jpg', tag_list: 'tag,2')
+    image3 = create_image(link: 'http://www.google.com/a.jpg', tag_list: '3')
```
rmacklin: Seems you could use `create_images` here

test/flows/images_crud_test.rb
```diff
@@ -63,7 +66,7 @@ class ImagesCrudTest < FlowTestCase
     puppy_url_1 = 'http://www.pawderosa.com/images/puppies.jpg'
     puppy_url_2 = 'http://ghk.h-cdn.co/assets/16/09/980x490/landscape-1457107485-gettyimages-512366437.jpg'
     cat_url = 'http://www.ugly-cat.com/ugly-cats/uglycat041.jpg'
-    Image.create!([
+    create_images([
       { link: puppy_url_1, tag_list: 'superman, cute' },
       { link: puppy_url_2, tag_list: 'cute, puppy' },
       { link: cat_url, tag_list: 'cat, ugly' }
```
rmacklin: Line 92 in this file (which I can't comment on directly) could also use `create_images`

test/controllers/images_controller_test.rb
lsq921224: Well I could, but since I am referring to the individual links I created, I did not do so
```
 assert_select 'img[src=?]', image1.link		 
 assert_select 'img[src=?]', image2.link
 assert_select 'img[src=?]', image3.link, count: 0
```

rmacklin: Are you familiar with ruby's parallel assignment feature/syntax?

You can do this:
```ruby
image1, image2, image3 = create_images(...)
```

lsq921224: well I didn't know this. It's good to know.

lsq921224:
I updated the code.

rmacklin:
👍

PR #52, opened 2016-09-14
Update rubocop setting to allow longer class length in app/controllers

lsq921224:
Update max class length to 150(was 100)

I had to increase the cap of class length because my images_controller is getting longer than 100 lines for 'share image part2 #50 '.


.rubocop.yml
```diff
@@ -26,3 +26,9 @@ Style/IndentHash:
 
 Style/MultilineMethodCallIndentation:
   EnforcedStyle: indented
+
+ClassLength:
+  Max: 200
```
rmacklin: This change will apply globally. Since you mentioned the only problem is in your `ImagesController` it would be much better to only change the limit there and keep the stricter limit everywhere else.

.rubocop.yml
```diff
@@ -26,3 +26,9 @@ Style/IndentHash:
...
+
+MethodLength:
+  Max: 15
```
rmacklin: Based on:
> I would like to increase the maximum method length as well because I think 10 is too short.

do you actually have methods that are hitting this limit? If not, I don't think we should raise the limit. If so, we should look at those methods first before deciding what to do.

.rubocop.yml
lsq921224: I would like to only apply the new maximum to `ImagesController`, but so far I only find a way to apply the new rule to all the controllers, i.e., put another `.rubocop.yml` in app/controllers that override the old one.

Otherwise, if I use something like `include: app/controllers/images_controller`, then that rule will only apply to that file, and rubocop won't check `class length` for all other files. 

If I use something like `exclude: app/controllers/images_controller`, then the rule simply won't check class length at all for images_controller.

So I wonder if there is a good way to only apply the new limit to `ImagesController`, but the old one to all other files?

rmacklin: Yeah, the other thing you can do is add special comments to tell rubocop to disable rules for specific blocks of code, but in this case that has the same drawback as:
> If I use something like `exclude: app/controllers/images_controller`, then the rule simply won't check class length at all for images_controller.

My suggestion is to bump the limit just for app/controllers, but not all the way to 200 if you don't need that many lines.

lsq921224:
I have updated this PR.

PR #55, opened 2016-09-15
Add style 2

lsq921224:
(Advanced) Push the footer to be on the bottom of the screen when there is little content.
(Advanced) Mark current page as active in the navbar.


lsq921224:
@rmacklin I was trying to use the following javascript to mark my current active `nav item`. The code works in dev and test, but cannot be precompiled on production.

I think it is the ES6 syntax that triggers the error
```
$(function() {
  let activeLink = $('.nav-link').filter(
    (index, link) => link.href == window.location
  );
  activeLink.parent().addClass('active');
});
```
> -----> Preparing app for Rails asset pipeline
>        Running: rake assets:precompile
>        rake aborted!
>        ExecJS::ProgramError: Unexpected token: name (activeLink) (line: 2, col: 6, pos: 21)
>        Error
>        at new JS_Parse_Error (/tmp/execjs20160915-1300-v615iojs:3572:11870)
>        at js_error (/tmp/execjs20160915-1300-v615iojs:3572:12089)

I see this gem in our project
```
# Allow use of ES6 (EMCAScript2015) modules.
gem 'browserify-rails'
```
and I thought it will take care of ES6?

rmacklin:
Browserify-rails is set up to transpile ES6, but browserify is only triggered if there is a `require` or `import` statement at the top level, which you don't have, so browserify isn't transforming it and it's falling back to sprockets.

Additionally, highlighting the current nav item can actually be done server side, and doing so would give you the opportunity to create your own helper that acts as a "pseudo" custom element. I could be wrong, but I think that's what @aaronpanch was going for with that exercise.

lsq921224:
I think it's better implement this feature in the client side, but I agree it would be good to practice

> doing so would give you the opportunity to create your own helper that acts as a "pseudo" custom element

I will see what @aaronpanch says

rmacklin:
Just curious, why do you think it's better? If JS is disabled, it won't work (which goes against what you did to make image sharing work even when JS is disabled).

lsq921224:
I think it's better to do this in the client side because this feature is entirely 'styling', and for styling I think it's better to delegate the job to the front end.
Also,` sharing image` is a 'functionality', and actually in this case the main functionality of this app, so i think it's good to make it work even when js is disabled. 
`Mark current page as active in the navbar.`, on the other hand, is not that important, and I think it's ok that it it does not work if js is disabled.

rmacklin:
> it's better to delegate the job to the front end.

The front end is not just JS - it's also HTML and CSS, which is what would be used to highlight the current page.

> Also,sharing image is a 'functionality', and actually in this case the main functionality of this app, so i think it's good to make it work even when js is disabled. 

👍 

> Mark current page as active in the navbar., on the other hand, is not that important, and I think it's ok that it it does not work if js is disabled.

That's a good stance. But at the same time if it _can_ be done without JS, it seems worth doing.

Anyway, as mentioned earlier, the bigger argument for doing it server side is from a learning perspective: it'd give you the opportunity to create a "pseudo" custom element, a la:
```erb
  <%= nav_bar ... %>
```

lsq921224:
>  create a "pseudo" custom element, a la:
> 
>   `<%= nav_bar ... %>`

In order to do this, do I need gems like `twitter-bootstrap-rails` ?
https://github.com/seyhunak/twitter-bootstrap-rails/wiki/Navbar-Helper

rmacklin:
> In order to do this, do I need gems like `twitter-bootstrap-rails` ?

No. You don't need any external dependencies to do this

"Pseudo custom elements" is not something I'd expect to return any relevant google results. "Custom Elements" on the other hand, should give results such as https://developers.google.com/web/fundamentals/primers/customelements/, though you don't need to read external resources to get the basic idea: Essentially, custom elements are an upcoming feature that will allow web developers to define their own custom HTML elements (and they can have associated javascript behavior if that's desired). Since not all browsers have support for this new feature yet, what I mean by "pseudo" custom elements is just ERB that looks like a custom element (e.g. `<%= nav_bar %>` as a "pseudo" custom element instead of a real `<nav-bar>` custom element), and this can be implemented by a Rails helper method that returns plain HTML for a nav bar.

From the link you shared it looks like `twitter-bootstrap-rails` does something similar, but the idea here was for you to see how you can implement these kinds of "pseudo" custom elements yourself.

lsq921224:
I have updated this PR but it fails to deploy on Heroku because of an error triggered by the SendGrid addon.
image

However, my other review apps are fine, so I am closing this PR and deleting this branch then pushing it again to see if that will work.

PR #28, opened 2016-08-04
Add style

lsq921224:
Closes #6

PR #57, opened 2016-09-21
Move flash_messages to layouts/ and use full variable name in it

lsq921224:
Addresses https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/28#pullrequestreview-1053988 and https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/28#pullrequestreview-1053512


app/views/shared/_flash_messages.html.erb
```diff
@@ -1,3 +0,0 @@
-<% flash.each do |name, msg| %>
-  <%= content_tag :div, msg, class: "alert alert-#{name}" %>
-<% end %>
```
rmacklin: You should move the file in a separate commit from the commit in which you modify the file. Otherwise, the history of the file will not be preserved, which is the case here. This is [Git Best Practice 2](https://sites.google.com/a/appfolio.com/eng/resources/our-tools/git)

app/views/layouts/application.html.erb
```diff
@@ -9,7 +9,7 @@
   <body>
     <%= render partial: 'layouts/header' %>
     <div class="container">
-      <%= render partial: 'shared/flash_messages' %>
+      <%= render partial: 'layouts/flash_messages' %>
```
rmacklin: Also, since you're making this change, it's going to have a conflict with https://github.com/AppFolioOnboarding/image-sharer-lsq921224/commit/d12b807a26a157bb4e4d0f046e315395af36fc14. That change could be applied on this branch and removed from the share image branch.

rmacklin:
The first two commits should be combined into one atomic commit to avoid breaking the application.

lsq921224:
Well I did that because in [Git Best Practice 2](https://sites.google.com/a/appfolio.com/eng/resources/our-tools/git), they separated them

> Here's an example of the right way:

> rename: https://github.com/appfolio/apm_bundle/commit/527097b41d115ef174575c14a7fd50694489a069
> 
> modify: https://github.com/appfolio/apm_bundle/commit/beeb3d3523d4d1e4cd2971ed86317b9c1a9b9691

PR #58, opened 2016-09-21
Refactor images controller

lsq921224:


app/controllers/images_controller.rb
```diff
@@ -1,5 +1,5 @@
 class ImagesController < ApplicationController
-  before_action :find_image, only: [:destroy, :share, :send_mail, :update]
+  before_action :find_image, except: [:new, :create, :index]
```
rmacklin: Prefer a whitelist (`only`) to a blacklist (`except`), as it requires an intentional change when a new action is added that _should_ use the `before_action`, rather than requiring an intentional change when a new action is added that _shouldn't_ use the `before_action`. That is, adding this extra behavior to actions should be "opt in" rather than "opt out"

app/controllers/images_controller.rb
```diff
@@ -86,5 +74,6 @@ def mail_form_params
 
   def find_image
     @image = Image.find_by(id: params[:id])
+    redirect_to images_path, error: 'Image not found. It may have been deleted.' unless @image.present?
```
rmacklin: Just curious: have you thought about how this change will affect your share image branch?

lsq921224: Yeah, I was going to add `send_mail` to the `except` list; now I just change this line to use `respond_to_no_image` instead, so my `send_mail` can use this `before_action` as well.

app/controllers/images_controller.rb
```diff
@@ -86,6 +73,9 @@ def mail_form_params
 
   def find_image
     @image = Image.find_by(id: params[:id])
+    return unless @image.nil?
+    flash[:error] = 'Image not found. It may have been deleted.'
```
rmacklin: Since the flash message is now generic, this line could be moved into the `respond_to_no_image` method

lsq921224:
I have updated this PR!

PR #59, opened 2016-09-22
Refactor code in preparation for share_image_part2

lsq921224:


lsq921224:
I recreated the `share_image_part2_improved_history` and I extracted the refactoring commits here, and the authors are me now!

PR #61, opened 2016-09-23
Refactor UserCreation test helper

lsq921224:
This is a very small PR that refactors UserCreation test helper:

it adds an extra number to name and email of the user created in `create_user`, so that `create_user` won't throw an error when being called multiple times

`create_user` also takes named arguments now


test/test_helpers/user_creation.rb
```diff
@@ -1,10 +1,10 @@
 module UserCreation
-  def create_user
+  def create_user(name: "foo#{User.count}", email: "foo#{User.count}@bar.com", password: 'password')
```
rmacklin: From the commit message:
> it adds an extra number to the user name and email of the user created, so that `create_user` won't throw an error when being called multiple times

This doesn't guarantee the method won't throw an error when called multiple times. Consider this scenario:
```ruby
first_user = create_user # creates User with email "foo0@bar.com"
second_user = create_user # creates User with email "foo1@bar.com"

first_user.destroy! # destroys User with email "foo0@bar.com"

# User.count is now 1, so:
create_user # tries to create User with email "foo1@bar.com", which throws an error
```

You could do something like:
```ruby
def create_user(name: nil, email: nil, password: 'password')
  last_user_id = User.last.try(:id) || 0
  name ||= "user#{last_user_id}"
  email ||= "#{name}@example.com"
  User.create!(...)
end
```
which still isn't foolproof but would take advantage of the autoincrement of the `id` column to attempt to avoid duplicates.

lsq921224: This is awesome. I have updated the PR to use this approach.

I wonder if there is a foolproof way to do this though? I think if I change the `id` to `current time stamp`, it will guarantee me a different user name every time?

rmacklin: There isn't a completely foolproof way because it's always possible that someone could have created a row in the DB with the value we want. That's probably less likely if you use a timestamp, but still possible. If you want to use a timestamp, that's fine with me, though using the autoincrement column is probably "good enough"

PR #63, opened 2016-09-28
Authorization

### LINKED ISSUE #14: Authorization
#### As an application hoster I want to restrict some functionality to specific users.

- [ ] Logged out users are restricted to read-only access.
- [ ] Links to non-read actions are not shown to logged out users.
- [ ] Attempting to access a page that requires authentication redirects you to
  the login page.
- [ ] Upon successful login, you are redirected back to where you had tried to
  navigate.
- [ ] Every image is associated with the user that added it.
- [ ] Only the user that added an image can delete that image.

Discussion topics:
- [ ] If we wanted to add a lot of authorization rules, how might we want to
  organize our code to keep it maintainable?

Further reading:
- https://github.com/elabs/pundit

Dependencies:
- Authentication

lsq921224:
Closes #14


jyoder:
In general, your commit messages should answer the question of "why?" was this change made. The "what?" change was made question is already answered by the code itself. Years later, some software engineer will come along and say, "why the heck did so-and-so make this change?" You want to help out future engineers (which may also be yourself in the future) :)

lsq921224:
I can totally agree with 

> Years later, some software engineer will come along and say, "why the heck did so-and-so make this change?" You want to help out future engineers (which may also be yourself in the future) :)

Right now my commit messages are basically a summary of what that commit does (I did include why I did that commit in some of the commit message's body, i.e

> 	Refactor log_in_from_index_page test helper  …
>            Allow log in as a specific user

 ).

I can change some of them to answer the 'why' as well, for example, 
`Add authorization on destroy action in images controller`
can be changed to 
`Add authorization for deleting an image`

but some of the commit's `why` is basically to make `authorization` work, for instance, my first two commits

> Integrate pundit to this app
> Add ApplicationPolicy

and

> Add ImagePolicy class used by pundit

should I just add things like `add authorization` or `preparation for authorization`to these commit messages?

jyoder:
In general, each commit should include the change you are making *and* the corresponding test for the change. Every change should have a corresponding test.

lsq921224:
thanks for the comment.

I have 
- reworded my commit messages so that they better answer the question of 'why'.
 
I also 
- added unit tests for `ApplicationPolicy` class in first commit https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/63/commits/1fc172da69e7f263688a7cbf58ae5bf2ea4e6719
- added unit tests for `ImagePolicy`class in second commit
https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/63/commits/4383eabec994d34decbee156ac0caacd7c1710a3
- added flow tests for testing `Show image buttons div only if use is logged in` for last commit https://github.com/AppFolioOnboarding/image-sharer-lsq921224/pull/63/commits/4fdd600bc8fff1f467b2217a99719fce1a9b9bdb

all the other commits either already have a corresponding test or do not really need one (I think)

test/test_helpers/flows/log_in_helper.rb
```diff
@@ -0,0 +1,34 @@
...
+      login_page.log_in!(
+        email: user.email,
+        password: user.password
+      ).as_a(PageObjects::Images::IndexPage)
+    end
+
+    def redirect_to_login_from(document, image: nil)
+      login_page = nil
+
+      Capybara.using_wait_time(0) do
```
jyoder: Just curious, what is the need for `using_wait_time(0)`?

app/controllers/sessions_controller.rb
```diff
@@ -25,6 +25,6 @@ def destroy
   def log_me_in(user)
     log_in user
     params[:session][:remember_me] == 'on' ? remember(user) : forget
-    redirect_to images_path, success: 'Logged in successfully.'
+    redirect_to session.delete(:last_url_navigation) || images_path, success: 'Logged in successfully.'
```
lsq921224: If I were to implement such class, where should I put it and how should I name it?

test/test_helpers/flows/log_in_helper.rb
lsq921224: I discussed this with Richard on slack, he said it's because we _know_ the page is going to redirect to the login page, so there's no point in letting the test wait up to capybara's default wait time (5 seconds) before catching the `visit` failure

lsq921224:
I have updated this PR in new commits.

app/controllers/sessions_controller.rb
jyoder: You could put it in `models`, or you could put it in `lib`. I'd call it something like `LoginRedirectService`. The interface could be something like:

`provider.has_redirect? => true|false`
`provider.update_redirect(url)`
`provider.consume_redirect => url`

test/test_helpers/flows/log_in_helper.rb
jyoder: 👍

app/policies/application_policy.rb
```diff
@@ -1,11 +1,13 @@
...
     @user = user
   end
 
   def logged_in?
     @user.present?
   end
+
+  private
+
+  attr_reader :user
```
jyoder: This is never called. You refer to `@user` in the class. I think you can just remove this accessor altogether.

test/controllers/images_controller_test.rb
```diff
@@ -345,4 +372,10 @@ def image_valid_params
       tag_list: 'tag, tag1, tag2'
     }
   end
+
+  def stub_log_in
+    user = create_user
+    ImagesController.any_instance.stubs(:current_user).returns(user)
```
rmacklin: Could this be changed to  `session[:user_id] = user.id` so we can avoid stubbing `current_user`? And I'd call this method `log_in` rather than `stub_log_in`

test/flows/images_crud_test.rb
```diff
@@ -162,12 +162,12 @@ def log_in_from_index_page
...
   end
 
-  def redirect_to_login_from(document)
+  def redirect_to_login_from(document, image: nil)
     login_page = nil
 
     Capybara.using_wait_time(0) do
       begin
-        document.visit
+        document.visit(image)
```
rmacklin: `image` only makes sense as a parameter to `visit` for a subset of pages in your application. This can/should be genericized to `*params`

app/views/images/_image.html.erb
```diff
@@ -18,7 +18,9 @@
                       image_url: image.link
                     } %>
         <%= link_to 'delete', image, method: :delete, data: { confirm: 'Are you sure?' }, class: 'btn btn-outline-danger pull-xs-right' %>
-        <%= link_to 'edit', edit_image_path(image), class: 'btn btn-outline-info pull-xs-right' %>
+        <% if policy(image).edit? %>
+          <%= link_to 'edit', edit_image_path(image), class: 'btn btn-outline-info pull-xs-right' %>
```
rmacklin: You should test that this link is present/absent when you expect it to be. You can do this using `assert_select` in your controller tests.

test/controllers/images_controller_test.rb
```diff
@@ -111,8 +111,10 @@ class ImagesControllerTest < ActionDispatch::IntegrationTest
     assert_select 'a[data-method="delete"]', count: 0
   end
 
-  test 'should display delete link if there are images' do
-    create_image
+  test 'should display delete button if there are images created by the (user logged in)' do
+    user = stub_log_in
+
+    create_image(user: user)
```
rmacklin: This test should create multiple images, not all of which are created by `user`, and then assert that the delete buttons only show for the images that were created by `user`

app/views/images/_image.html.erb
```diff
@@ -10,13 +10,15 @@
...
       <div class="image-card__buttons">
-        <%= link_to 'share', share_image_path(image), class: 'btn btn-primary',
-                    data: {
-                      toggle: 'modal',
-                      target: '#share_modal',
-                      image_id: image.id,
-                      image_url: image.link
-                    } %>
+        <% if policy(image).share? %>
+          <%= link_to 'share', share_image_path(image), class: 'btn btn-primary',
```
rmacklin: You should test that this link is present/absent when you expect it to be. You can do this using `assert_select` in your controller tests.

app/policies/image_policy.rb
```diff
@@ -34,6 +34,10 @@ def send_mail?
     logged_in?
   end
 
+  def viewer?
```
rmacklin: This method is poorly named. Additionally, what is the point of the `viewer?` method over just using `logged_in?` directly (which has a clear meaning)?

test/flows/images_flow_test.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  test 'do not show buttons div below images card if user is not logged in' do
+    create_image
+
+    images_index_page = PageObjects::Images::IndexPage.visit
+
+    image_card = images_index_page.images.first
+    assert image_card.buttons_div.absent?
+  end
+end
```
rmacklin: This stuff can be checked in a (much faster) controller test.

test/page_objects/images/image_card.rb
```diff
@@ -1,6 +1,8 @@
 module PageObjects
   module Images
     class ImageCard < AePageObjects::Element
+      element :buttons_div, locator: '.image-card__buttons'
```
rmacklin: `div` is insignificant in the page object interface. This element represents "image actions". so a more appropriate name would be `image_actions`.

rmacklin:
153b28964db3f29595ee4fe2fd76a0b676d82970 is not atomic. Adding the migration and updating the model and seeds is its own change, separate from adding authorization for image creation

rmacklin:
Regarding the acceptance criterion:
> Links to non-read actions are not shown to logged out users.

I can see the New Image link even when logged out

app/policies/image_policy.rb
```diff
@@ -0,0 +1,46 @@
...
+    super(user)
+    @image = image
+  end
+
+  def new?
+    logged_in?
+  end
+
+  def create?
+    logged_in?
```
rmacklin: You want this to have the same permissions as `new?`, so it's better if this calls `new?` rather than repeating the same logic as `new?` (which currently isn't much, but if it changes, we should only have to change _one_ place, not two).

app/policies/image_policy.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def edit?
+    logged_in?
+  end
+
+  def update?
+    logged_in?
```
rmacklin: You want this to have the same permissions as `edit?`, so it's better if this calls `edit?` rather than repeating the same logic as `new?` (which currently isn't much, but if it changes, we should only have to change _one_ place, not two).

app/policies/image_policy.rb
```diff
@@ -0,0 +1,46 @@
...
+  def destroy?
+    logged_in? && owner?
+  end
+
+  def share?
+    logged_in?
+  end
+
+  def send_mail?
+    logged_in?
```
rmacklin: You want this to have the same permissions as `share?`, so it's better if this calls `share?` rather than repeating the same logic as `new?` (which currently isn't much, but if it changes, we should only have to change _one_ place, not two).

app/policies/image_policy.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def viewer?
+    logged_in?
+  end
+
+  private
+
+  def owner?
```
rmacklin: `user_owns_image?` would be a bit more explicit/descriptive.

test/controllers/images_controller_test.rb
```diff
@@ -1,10 +1,21 @@
...
   include ImageCreation
+  include UserCreation
+
+  test 'redirect user to log in page if user is not logged in and visits new page' do
+    get new_image_path
+
+    assert_redirected_to new_session_path
+    assert_equal 'You need to log in first.', flash[:error]
+  end
+
```
rmacklin: The `create`, `update`, and `send_mail` actions are subject to the same "redirect if unauthorized" logic, so there should be corresponding tests validating that behavior for each of those actions (in addition to the tests for their counterparts `new`, `edit`, and `share`)

